PR #12, opened 2021-11-10
Add skip property to permission schema to be able to skip placeholder.

OliviaHY:
### Why&What
To be able to skip placeholder. Specifically for slakc-usergroup integration in this case, therefore updating the PermissionSlackUsergroup_v1 schema.


maorfr:
Perhaps add `skip` under the slack onion to make sure it's only used where we expect?

Piojo (approved):
Address @maorfr's comment before pushing.

maorfr (approved):
LGTM

PR #37, opened 2021-12-16
Add Add support for SNS event notification for S3 and consolidate it with SQS

OliviaHY:
### Why & What:
It is requested in [APPSRE-3780](https://issues.redhat.com/browse/APPSRE-3780) by to add support for s3://insights-warehouse-prod and s3://insights-ingress-prod buckets to able to publish to a couple of SNS topics via arn.

This schema update will allow adding event notification as an array that configure SNS or SQS event notification to the bucket.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -749,6 +749,14 @@
   - { name: description, type: string }
   - { name: region, type: string, isRequired: true }
 
+- name: AWSS3EventNotification_v1
+  fields:
+  - { name: event_type, type: string, isRequired: true, isList: true}
+  - { name: destination, type: string, isRequired: true}
+  - { name: destination_type, type: string, isRequired: true}
+  - { name: filters_prefix, type: string}
+  - { name: filters_suffix, type: string}
+
 - name: ACMDomain_v1
   fields:
   - { name: domain_name, type: string, isRequired: true }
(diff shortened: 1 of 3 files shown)


maorfr:
This looks pretty good!
I do think there are 2 features here: sns notifications and sns support. Let's start with only notifications and split out the sns support to a separate PR (and even ticket).

In the ticket you are working on (please add a link to it in the description) we only need notifications (the sns topic is created out of our tools).

For notifications, I would rather see something like this schema:
```
notifications:
  sqs_identifier: ...
  sns_identifier: ...
```

So it would probably make sense to start with only changing the sqs (before adding support for sns). This will be a good exercise in schema migration and will allow easier reviews and faster iterations.

WDYT?

OliviaHY:
> 
Thank you for the review! So I'm working on the schema update like you describe to manage event notifications as a whole for both SNS and SQS.
My plan was to get the functionality for both supporting the SNS notification and SNS resource management. The reason being I want to get something out for them to use sooner then later, but I have no problem reorganize the work like you asked.

maorfr (approved):
LGTM!

Thanks for splitting this up!

As this is a breaking change, perhaps you'd prefer to keep the existing fields (backwards compatibility) so you are not in a rush to update all the data and logic. Leaving it for you to decide.

maorfr (approved):
LGTM!!!

PR #40, opened 2021-12-20
Add per-path targets for ALB

syed:
ref: https://github.com/app-sre/qontract-reconcile/pull/2057
JIRA: APPSRE-3695

This schema change adds per-path targets instead of read/write targets. A sample config looks something like this:

```
provider: alb
...
targets:
- name: py2
  openshift_service: py2-svc
  default: true
- name: py3
  openshift_service: py3-svc
rules:
- condition:
    path: /v2/auth*
    methods: read
  action:
  - target: py2
    weight: 100
- condition:
    path: /v2*
    methods: read
  action:
  - target: py3
    weight: 100
  - target: py2
    weight: 0
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1113,7 +1113,7 @@
   - { name: certificate_arn, type: string, isRequired: true }
   - { name: idle_timeout, type: int }
   - { name: targets, type: NamespaceTerraformResourceALBTargets_v1, isRequired: true, isList: true }
-  - { name: paths, type: NamespaceTerraformResourceALBPaths_v1 }
+  - { name: rules, type: NamespaceTerraformResourceALBRules_v1, isRequired: true, isList: true }
   - { name: vpc, type: AWSVPC_v1, isRequired: true }
   - { name: output_resource_name, type: string }
   - { name: annotations, type: json }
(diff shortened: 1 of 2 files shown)


schemas/openshift/terraform-resource-1.yml
```diff
@@ -635,24 +627,44 @@ oneOf:
...
+        additionalProperties: false
+        properties:
+          condition:
+            type: object
+            additionalProperties: false
+            properties:
+              path:
+                type: string
+              methods:
+                type: string
```
Piojo, line 648: Shouldn't this one be also an `enum`?

syed: This was originally a `string` so I kept it that way. I can change it to an `enum`

graphql-schemas/schema.yml
```diff
@@ -1092,19 +1092,23 @@
...
   - { name: ips, type: string, isList: true }
   - { name: openshift_service, type: string }
 
-- name: NamespaceTerraformResourceALBTargetWeights_v1
+- name: NamespaceTerraformResourceALBRules_v1
   fields:
-  - { name: read, type: int, isRequired: true }
-  - { name: write, type: int, isRequired: true }
+  - { name: condition: type: NamespaceTerraformResourceALBConditon_v1, isRequired: true}
+  - { name: action: type: NamespaceTerraformResourceALBAction_v1, isRequired: true, isList: true}
```
jfchevrette, line 1101:
```suggestion
  - { name: condition, type: NamespaceTerraformResourceALBConditon_v1, isRequired: true}
  - { name: action, type: NamespaceTerraformResourceALBAction_v1, isRequired: true, isList: true}
```
fix typo

graphql-schemas/schema.yml
```diff
@@ -1092,19 +1092,23 @@
...
   fields:
-  - { name: read, type: string, isList: true }
-  - { name: write, type: string, isList: true }
+  - { name: path: type: string, isRequired: true}
+  - { name: methods: type: string, isRequired: true}
+
+- name: NamespaceTerraformResourceALBAction_v1
+  fields:
+  - { name: target: type: string, isRequired: true}
+  - { name: weight: type: int, isRequired: true}
```
jfchevrette, line 1111:
```suggestion
  - { name: path, type: string, isRequired: true}
  - { name: methods, type: string, isRequired: true}

- name: NamespaceTerraformResourceALBAction_v1
  fields:
  - { name: target, type: string, isRequired: true}
  - { name: weight, type: int, isRequired: true}
```

more `:`s instead of `,`s :)

graphql-schemas/schema.yml
```diff
@@ -1122,19 +1122,23 @@
...
   - { name: ips, type: string, isList: true }
   - { name: openshift_service, type: string }
 
-- name: NamespaceTerraformResourceALBTargetWeights_v1
+- name: NamespaceTerraformResourceALBRules_v1
   fields:
-  - { name: read, type: int, isRequired: true }
-  - { name: write, type: int, isRequired: true }
+  - { name: condition, type: NamespaceTerraformResourceALBConditon_v1, isRequired: true}
+  - { name: action, type: NamespaceTerraformResourceALBAction_v1, isRequired: true, isList: true}
```
jfchevrette, line 1131: AWS seem to support multiple conditions and a single action per rule. I think perhaps you meant to set `isList` to the condition element here?

jfchevrette: Thinking about this some more I think the current implementation is fine for the purpose of specifying both a path and methods once (vs terraform or aws implementing this as multiple conditions, that can be many things) and a list of actions in our case as we only support for forward action kind which itself supports multiple targets & weights. This is effectively abstracting and hardcoding the forward actions which is fine here

All that to say that in the end I think we don't need to change this until we may want to support other kinds of complex configurations in the future

PR #45, opened 2022-01-06
[endpoint monitoring] Schema enhancement for endpoint monitoring with blackbox-exporter

geoberle:
Schema changes for blackbox-exporter based service endpoint monitoring

Ref: https://issues.redhat.com/browse/APPSRE-4161

### DIFF
--- a/schemas/dependencies/endpoint-monitoring-provider-1.yml
+++ b/schemas/dependencies/endpoint-monitoring-provider-1.yml
@@ -20,23 +20,31 @@ properties:
     type: string
     enum:
     - blackbox-exporter
+    description: The provider defines which monitoring implementation will be used for endpoints
   timeout:
     type: string
+    description: Timeout for the monitoring check, e.g. 10s
   checkInterval:
     type: string
+    description: Interval between endpoint checks, e.g. 10s
   metricLabels:
     "$ref": "/common-1.json#/definitions/labels"
+    description: Labels to be added to the resulting Prometheus metrics
   blackboxExporter:
       type: object
+      description: blackbox-exporter specific configuration options
       additionalProperties: false
       properties:
         module:
           type: string
+          description: the blackbox exporter module to use for the endpoint check, e.g. http_2xx
         namespace:
           "$ref": "/common-1.json#/definitions/crossref"
           "$schemaRef": "/openshift/namespace-1.yml"
+          description: the namespace where the prometheus Probe CR will be created
         exporterUrl:
           type: string
+          description: the blackbox-exporter URL to use, must be a full URL including potential URL paths
       required:
       - module
       - namespace
--- a/schemas/dependencies/endpoint-monitoring-provider-1.yml
+++ b/schemas/dependencies/endpoint-monitoring-provider-1.yml
@@ -1,0 +1,71 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /dependencies/endpoint-monitoring-provider-1.yml
+  labels:
+    "$ref": "/common-1.json#/definitions/labels"
+  name:
+    type: string
+  description:
+    type: string
+  provider:
+    type: string
+    enum:
+    - blackbox-exporter
+  timeout:
+    type: string
+  checkInterval:
+    type: string
+  metricLabels:
+    "$ref": "/common-1.json#/definitions/labels"
+  blackboxExporter:
+      type: object
+      additionalProperties: false
+      properties:
+        module:
+          type: string
+        namespace:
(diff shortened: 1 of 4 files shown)


rporres:
Looks great, as the companion PR. I would add more descriptions to document the different fields for those that go to the schema to look for answers 😂

geoberle:
> Looks great, as the companion PR. I would add more descriptions to document the different fields for those that go to the schema to look for answers 😂 

Done

schemas/dependencies/endpoint-monitoring-provider-1.yml
```diff
@@ -0,0 +1,79 @@
...
+  description:
+    type: string
+  provider:
+    type: string
+    enum:
+    - blackbox-exporter
+    description: The provider defines which monitoring implementation will be used for endpoints
+  timeout:
+    type: string
+    description: Timeout for the monitoring check, e.g. 10s
```
rporres, line 26: Should this have a regex so that it is clearer what it is accepted, e.g. `^\d+[smhd]$`?

schemas/dependencies/endpoint-monitoring-provider-1.yml
```diff
@@ -0,0 +1,79 @@
...
+  checkInterval:
+    type: string
+    description: Interval between endpoint checks, e.g. 10s
```
rporres, line 29: Should this have a regex so that it is clearer what it is accepted, e.g. `^\d+[smhd]$`?

schemas/dependencies/endpoint-monitoring-provider-1.yml
```diff
@@ -0,0 +1,79 @@
...
+  metricLabels:
+    "$ref": "/common-1.json#/definitions/labels"
+    description: Labels to be added to the resulting Prometheus metrics
```
rporres, line 32: Prometheus label names cannot have any character, as `string` provides. From https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels they must match `[a-zA-Z_][a-zA-Z0-9_]*`

schemas/dependencies/endpoint-monitoring-provider-1.yml
```diff
@@ -0,0 +1,79 @@
...
+  blackboxExporter:
+      type: object
+      description: blackbox-exporter specific configuration options
+      additionalProperties: false
+      properties:
+        module:
+          type: string
+          description: the blackbox exporter module to use for the endpoint check, e.g. http_2xx
```
rporres, line 40: Should `module` be an enum, since we know which modules there are?

rporres (commented):
I took a new look into this. Thx a lot for the descriptions, it made things much easier to grasp.

I think we should be tighter in the values accepted to avoid late minute surprises.

schemas/dependencies/endpoint-monitoring-provider-1.yml
geoberle: fixed in c209757

schemas/dependencies/endpoint-monitoring-provider-1.yml
geoberle: fixed in c209757

schemas/dependencies/endpoint-monitoring-provider-1.yml
geoberle: fixed in c209757

schemas/dependencies/endpoint-monitoring-provider-1.yml
geoberle: the list of modules in blackbox_exporter is not static. additional modules can be defined in, see [here](https://github.com/prometheus/blackbox_exporter#configuration). http_2xx is one of the predefined ones

rporres: Indeed, but we control blackbox exporter, we know which modules are available. What I'm trying to avoid here is a typo in the module name not being caught at a MR level in app-interface and then making things fail much later. I prefer to maintain a list of what we accept here than facing the potential debugging of that

geoberle: i thought that things too specific for AppSRE should be avoided in schemas in the spirit of opensource.

rporres: That's a fair point. Then the only chance we have is to add the enumeration in an app-interface setting that is checked from the integration so that we have the best of both worlds :)

rporres: I don't want to block on this. If you agree with the solution, you can totally take care of it in a subsequent PR

geoberle: agreed. thank you for your thorough review!

rporres (approved):

PR #46, opened 2022-01-06
Correct redundant plural.

OliviaHY:
This was not consistent with the [terraform resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_notification)


graphql-schemas/schema.yml
```diff
@@ -755,8 +755,8 @@
   - { name: event_type, type: string, isRequired: true, isList: true}
   - { name: destination, type: string, isRequired: true}
   - { name: destination_type, type: string, isRequired: true}
-  - { name: filters_prefix, type: string}
```
OliviaHY, line 758: I added this schema recently and it is not being used yet in App Interface.

TGPSKI (approved):

PR #49, opened 2022-01-17
Make tekton provider defaults mandatory

rporres:
We will make sure that all pipelines providers always have what they need
to work properly and will fail at a schema level without the need of any
ad-hoc checks in the the code.

This is a breaking change.

part of APPSRE-4165


andreugallofre:
LGTM

maorfr (approved):

PR #52, opened 2022-01-18
APPSRE-4229 cluster network type support

patjlm:
New optional enum field `type` within `ClusterNetwork_v1`.
Default will be `OpenShiftSDN`, handled in qontract-reconcile: https://github.com/app-sre/qontract-reconcile/pull/2099


geoberle (approved):
LGTM

maorfr (approved):

PR #53, opened 2022-01-18
add image_data

BumbleFeng:
Part of https://issues.redhat.com/browse/APPSRE-3925

Design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34670

### DIFF
--- a/schemas/openshift/terraform-resource-1.yml
+++ b/schemas/openshift/terraform-resource-1.yml
@@ -139,12 +139,6 @@ properties:
       type: string
   image:
     type: object
-    additionalProperties: false
-    properties:
-      id:
-        type: string
-    required:
-    - id
 oneOf:
 - "$ref": "/aws/s3-1.yml"
 - additionalProperties: false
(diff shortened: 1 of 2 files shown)


schemas/openshift/terraform-resource-1.yml
```diff
@@ -794,8 +788,27 @@ oneOf:
...
+            tag_name:
+              type: string
+            url:
+              type: string
+              format: uri
+            ref:
+              type: string
+              pattern: '^([0-9a-f]{40}|master|main)$'
+          required:
+          - tag_name
```
geoberle, line 804: i understand ref and url, but what is tag_name?

BumbleFeng: We assume AMI should be tagged with a tag_name and its value should be commit_sha comes from url and ref. So we can filter AMI with that tag. tag_name can be customized by tenants(e.g image builder use composer_commit)

rporres:
Can you add documentation using `description` field to all the new schema elements?

schemas/openshift/terraform-resource-1.yml
```diff
@@ -794,8 +788,27 @@ oneOf:
       properties:
         id:
           type: string
```
maorfr, line 790: as discussed, image id should be removed

maorfr (approved):

PR #54, opened 2022-01-19
dyn-traffic-director: set range for record weight

jfchevrette:
Dyn requires a weight of  `1..255` so this limits the schema such that validation will fail early


jfchevrette:
Tested locally

maorfr (approved):

PR #55, opened 2022-01-20
Add tests for SQL query regular expression

lechuk47:
## Overview 
This PR adds test cases to ensure the sql-query regex validates the queries allowed in by the schema. The code also introduces a helper function to get validators for any other object that might need testing.
This is part of APPSRE-3316 to improve the sql queries regexp to allow comments on them. 

## Notes: 
* Running these tests here instead of in app-interface it's faster and it provides a more direct way to test the expressions.
* A possible approach is to generate fake data and use qontract-validator, but it introduces a lot of boilerplate files/data.

### DIFF
--- a/test/test_sql_regexp.py
+++ b/test/test_sql_regexp.py
@@ -1,0 +1,53 @@
+import pytest
+import anymarkup
+from typing import Any, Dict
+from jsonschema.exceptions import ValidationError
+
+from utils import get_schema_object_validator
+
+valid_sql_queries = [
+    "select a from table_a;",
+    "SELECT A FROM TABLE_A;",
+    "select a from table order by id;",
+    "select a from table order by id;",
+    "select a from ( select a,b from table_b );",
+    "explain select a from table_a;"
+]
+
+not_valid_sql_queries = [
+    "select * from table_a",
+    "SELECT * FROM TABLE_A",
+    "select a from ( select * from table_b )",
+    "explain analyze select a from table b"
+]
+
+def new_query_spec(query: str) -> Dict[str, Any]:
+    return {
+        "$schema": "/app-interface/app-interface-sql-query-1.yml",
+        "labels": {},
+        "name": "query",
+        "namespace": {
+            "$ref": "namespace.yaml"
+        },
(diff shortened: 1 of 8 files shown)


test/test_sql_regexp.py
```diff
@@ -0,0 +1,53 @@
...
+from typing import Any, Dict
+from jsonschema.exceptions import ValidationError
+
+from utils import get_schema_object_validator
+
+valid_sql_queries = [
+    "select a from table_a;",
+    "SELECT A FROM TABLE_A;",
+    "select a from table order by id;",
+    "select a from table order by id;",
```
steveteahan, line 12: nit: is this the same as the line above? Was it meant to be a multi-column query, or something else along those lines?

steveteahan (approved):
Nice use of pytest fixtures and parameterization!

test/test_sql_regexp.py
```diff
@@ -0,0 +1,53 @@
...
+        "name": "query",
+        "namespace": {
+            "$ref": "namespace.yaml"
+        },
+        "identifier": "query",
+        "output": "stdout",
+        "query": query,
+    }
+
+@pytest.fixture(scope='session')
```
rporres, line 37: Just to verify that I'm understanding this correctly. The `session` scope actually means that this fixture is only created once in the whole test execution so that `load_schemas` is called just once?

test/test_sql_regexp.py
lechuk47: it was intended to be a query with a comment, I removed the comment to just pass the tests bc the regex is not updated yet.

test/test_sql_regexp.py
lechuk47: Yes. Your comment has made me change the code. I think it's better to have a session scoped fixture that loads the schemas (only happens once in the pytest run) and have another module scoped fixture to get the sql validator. 

I think it makes now and is more reusable for future tests.

rporres: Yeah, I like it more this way, it is definitely more reusable.

rporres (approved):

PR #56, opened 2022-01-24
Added AWS ES advanced security options support

rporres:
part of [APPSRE-4219](https://issues.redhat.com/browse/APPSRE-4219)

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elasticsearch_domain#advanced_security_options

### DIFF
--- a/schemas/aws/elasticsearch-defaults-1.yml
+++ b/schemas/aws/elasticsearch-defaults-1.yml
@@ -102,5 +102,32 @@ properties:
         type: string
       indices.query.bool.max_clause_count:
         type: string
+  advanced_security_options:
+    type: object
+    additionalProperties: false
+    properties:
+      enabled:
+        type: boolean
+      internal_user_database_enabled:
+        type: boolean
+      master_user_options:
+        type: object
+        additionalProperties: false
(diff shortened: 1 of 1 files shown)


schemas/aws/elasticsearch-defaults-1.yml
```diff
@@ -102,5 +102,32 @@ properties:
...
         type: string
+  advanced_security_options:
+    type: object
+    additionalProperties: false
+    properties:
+      enabled:
+        type: boolean
+      internal_user_database_enabled:
+        type: boolean
+      master_user_options:
```
Piojo, line 113: Mmm... if everything related to this user is going to be in vault, do we need the extra level of indentation?

Piojo (approved):
One question, but it does match the Terraform documentation.

schemas/aws/elasticsearch-defaults-1.yml
rporres: I want to match the structure of the original terraform config as close as possible to make things easier... An in the future we may want to add the other auth option (via IAM) that AWS supports

PR #58, opened 2022-01-24
Improve sql query validator pattern

lechuk47:
This PR improves the sql validator pattern adding the following capabilities. 
Jira: APPSRE-3316

* Add comments to queries with the `/* [comment] */` format. `--` at the end of line is not supported 
* `explain analyze select` is now supported. A tenant might require to use this statement to evaluate costs, etc. 
  The regexp restricts the usage of `explain analyze` to use just `select` statements

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -5,7 +5,7 @@
   "definitions": {
     "query": {
       "type": "string",
-      "pattern": "(?i)^(select|explain) (?!analyze (?!select))[^;\\*]+;$"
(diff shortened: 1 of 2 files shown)


Piojo (approved):

rporres:
I got late here...

I'm wondering if we need the `.*?` in some of the places. I was about to test something like this:

`(?i)^(select|explain(?!\\s+?analyze\\s+(insert|update|delete)))\\s+((\\/\\*.*\\*\\/)|[^;\\*-])+;`

* Which removes the `.*?`
* Removes the non-capturing group to make things more consistent

PR #60, opened 2022-01-24
vault: add totp, make options optional

jfchevrette:
part of [APPSRE-4273](https://issues.redhat.com/browse/APPSRE-4273)

https://www.vaultproject.io/docs/secrets/totp

This adds the TOTP engine as an engine type. TOTP has no options so we now make options optional


maorfr (approved):
LGTM

PR #62, opened 2022-01-25
Improve SQL query regexp

lechuk47:
## Changes made: 

* Regex now accepts comments in both formats `-- comment \n` and `/* comment */`
* `*` is now allowed in comments and in values, e.g: `where id='an_id_with_*'`

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -5,7 +5,7 @@
   "definitions": {
     "query": {
       "type": "string",
-      "pattern": "(?i)^(select|explain) (?!analyze (?!select))[^;\\*]+;$"
+      "pattern": "(?i)^(select|explain(?!\\s+analyze\\s+(insert|update|delete)))\\s+((?:\\/\\*.*?\\*\\/)|[^;\\*-])+;$"
     },
     "labels": {
       "type": "object",
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -5,7 +5,7 @@
   "definitions": {
     "query": {
       "type": "string",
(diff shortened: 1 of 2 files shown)


schemas/common-1.json
```diff
@@ -5,7 +5,7 @@
   "definitions": {
     "query": {
       "type": "string",
-      "pattern": "(?i)^(select|explain) (?!analyze (?!select))[^;\\*]+;$"
+      "pattern": "(?i)^(select|explain(?!.*analyze.*(insert|update|delete)))\\s+((?:\\/\\*.*\\*\\/)|(?:[\\'\\\"].*[\\'\\\"])|[^;\\*-])+;$"
```
rporres, line 8: why the non-capturing groups in the second part of the regex?

rporres:
Good stuff, great work! Being a regex far from trivial and not being able to add comments in a multiline regexp, how about throwing somewhere some notes about it? Unfortunately the schema file is in json, so it doesn't accept comments (and the `$comment` keyword was only added in Draft 7) so I don't have a good idea about where to add the notes

lechuk47:
@rporres please take a look now. 
I've changed the thing a bit as there are some cases where the previous expression was not working. Basically now, it supports both type of comments and '*' is possible to be used in values. 

There are pretty complex queries in app-interface and I think a regexp is not the best approach to validate the syntax, so I think it's better to just focus this effort to add support for comments and not check the sql syntax in any way.

rporres (approved):
nit: I'd be a bit more consistent with non-capturing groups, but it's not extremely important. The main thing is what you have accomplished here. Good stuff!

PR #63, opened 2022-01-25
APPSRE-4119 add publish_log_types list for ES

fishi0x01:
This adds `publish_log_types` to the schema. The types will be consumed in a next step by our terraform integration to forward logs to cloudwatch loggroups accordingly.

Usage example:
```
terraformResources:
- provider: elasticsearch
  account: xxx
  identifier: kibana-elasticsearch
  output_resource_name: kibana-elasticsearch
  publish_log_types:
  - ES_APPLICATION_LOGS
  - SEARCH_SLOW_LOGS
```

Tested locally by adding following to one of our existing services in app-interface:
```
- provider: elasticsearch
  ...
  publish_log_types:
  - ES_APPLICATION_LOGS
  - SEARCH_SLOW_LOGS
```

Then run:
```
app-sre/qontract-server$ make dev 
...
WARNING: We can't validate resource with schema ...
[]
yarn run v1.22.17
$ node ./dist/main-bundle.js
2022-01-25T17:17:57.599Z INFO: Running at http://localhost:4000/graphql
```

Further, checked in http://localhost:4000/graphql that the terrascript query returns `publish_log_types` properly.

### DIFF
--- a/schemas/openshift/terraform-resource-1.yml
+++ b/schemas/openshift/terraform-resource-1.yml
@@ -540,7 +540,6 @@ oneOf:
         - INDEX_SLOW_LOGS
         - SEARCH_SLOW_LOGS
         - ES_APPLICATION_LOGS
-        - AUDIT_LOGS
   required:
   - account
   - identifier
--- a/schemas/openshift/terraform-resource-1.yml
+++ b/schemas/openshift/terraform-resource-1.yml
@@ -540,6 +540,7 @@ oneOf:
         - INDEX_SLOW_LOGS
         - SEARCH_SLOW_LOGS
         - ES_APPLICATION_LOGS
+        - AUDIT_LOGS
   required:
   - account
   - identifier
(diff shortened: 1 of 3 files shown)


schemas/openshift/terraform-resource-1.yml
```diff
@@ -131,6 +131,15 @@ properties:
     type: array
     items:
       type: object
+  publish_log_types:
+    type: array
+    items:
+      type: string
+      enum:
```
maorfr, line 138: you can probably leave out the enum in this section to avoid duplication?

fishi0x01: I thought so too, but sadly it doesn't work if I leave this out:

```
"reason": "VALIDATION_ERROR",
"error": "Additional properties are not allowed ('publish_log_types' was unexpected)
```

maorfr: I meant only the enum, not the entire thing

fishi0x01: Good idea!
Fixed

maorfr (approved):

PR #66, opened 2022-01-28
Grafana Urls as list

janboll:
Make it possible to display multiple Grafana Urls for a given service

This implements requirements for configuring multiple dashboards, as described in this ticket: https://issues.redhat.com/browse/APPSRE-3494

Corresponding update in visual qontract: https://github.com/app-sre/visual-qontract/pull/149

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -236,6 +236,11 @@
   - { name: automationToken, type: VaultSecret_v1 }
   - { name: pushCredentials, type: VaultSecret_v1 }
 
+- name: GrafanaDashboardUrls_v1
+  fields:
+  - { name: title, type: string, isRequired: true}
+  - { name: url, type: string, isRequired: true}
+
 - name: QuayOrg_v1
   fields:
   - { name: schema, type: string, isRequired: true }
@@ -1403,7 +1408,8 @@
   - { name: name, type: string, isRequired: true }
   - { name: description, type: string, isRequired: true }
   - { name: onboardingStatus, type: string, isRequired: true }
-  - { name: grafanaUrl, type: string, isRequired: true }
+  - { name: grafanaUrl, type: string }
+  - { name: grafanaUrls, isList: true, type: GrafanaDashboardUrls_v1 }
   - { name: sopsUrl, type: string, isRequired: true }
   - { name: architectureDocument, type: string, isRequired: true }
   - { name: parentApp, type: App_v1 }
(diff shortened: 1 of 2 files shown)


schemas/app-sre/app-1.yml
```diff
@@ -27,9 +27,20 @@ properties:
...
+      a title together with the link.
+    items:
+      type: object
+      properties:
+        title:
+          type: string
+        url:
+          type: string
+          format: uri
+      minItems: 1
```
rporres, line 43: In order to make this object a bit more tight schema wise, I would add
```yaml
additionalProperties: false
required:

**Comment by rporres** [comment]
This is a breaking change.

cc @maorfr @TGPSKI

**Comment by rporres** [comment]
@janboll Thinking about how you're going to change data and the review process, you may want to split this change in two:

1. Add the `grafanaUrls` object and remove the requirement for `grafanaUrl`
2. Make `grafanaUrls` mandatory

The rationale is to change in things in a way that you can change data in `app-interface` between 1. and 2. using a schema version that doesn't break the existing data.

**Comment by maorfr** [comment]
proceed as you fit, we have very little data in fedramp, so migration is ezpz.

File: schemas/app-sre/app-1.yml (hunk 1 above)
Comment (janboll) [reply schemas/app-sre/app-1.yml:51]: :+1:

**Comment by janboll** [comment]
> @janboll Thinking about how you're going to change data and the review process, you may want to split this change in two:
> 
> 1. Add the `grafanaUrls` object and remove the requirement for `grafanaUrl`
> 2. Make `grafanaUrls` mandatory
> 
> The rationale is to change in things in a way that you can change data in `app-interface` between 1. and 2. using a schema version that doesn't break the existing data.

Great idea. I added the `grafanaUrl` back to the PR.

File: graphql-schemas/schema.yml [hunk 2]
```diff
@@ -1403,7 +1408,8 @@
   - { name: name, type: string, isRequired: true }
   - { name: description, type: string, isRequired: true }
   - { name: onboardingStatus, type: string, isRequired: true }
-  - { name: grafanaUrl, type: string, isRequired: true }
+  - { name: grafanaUrls, isList: true, type: GrafanaDashboardUrls_v1 }
+  - { name: grafanaUrls, isList: true, type: GrafanaDashboardUrls_v1 }
```
rporres, line 1412: Duplicated property. I would leave grafanaUrl until you remover ir completely

janboll: :+1:

rporres (approved):

PR #69, opened 2022-02-01
Update user-1.yml schema to restrict aws username pattern

TGPSKI:
This MR restricts `aws_username` field values to lowercase letters and three word blocks separated by underscores.

i.e.`tpate_rh_us`


maorfr (approved):
this is very specific to our use case. would it make sense to allow low case characters and underscores and not define the specific `a_b_c` structure?

in any case, LGTM. feel free to choose the right implementation

PR #76, opened 2022-02-10
APPSRE-4339 AWS uid must not be unique

fishi0x01:
**Motivation**

We want to simulate multiple accounts in test runs. For that reason, we must be able to allow multiple account aliases to point to the same AWS `uid`.

Decision for this change: https://issues.redhat.com/browse/APPSRE-4399

**Changes**

Remove uniqueness constraint from `uid`

**Test**

Successfully bundled with schema multiple accounts in app-interface locally.


BumbleFeng:
lgtm

BumbleFeng (approved):

PR #77, opened 2022-02-14
add description field to VPC peering connections

geoberle:
the purpose for VPC peerings is not always obvious. this change
introduces optional description fields to allow documentation
of VPC peering connections.

Ref: https://issues.redhat.com/browse/APPSRE-4428


patjlm:
LGTM

PR #81, opened 2022-02-21
add support to define aws policy as mandatory

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3251

with this change, we can implement a validation in qontract-reconcile.

### DIFF
--- a/schemas/aws/policy-1.yml
+++ b/schemas/aws/policy-1.yml
@@ -20,6 +20,7 @@ properties:
     type: string
   mandatory:
     type: boolean
+    description: is policy mandatory in the account
   policy:
     type: object
 required:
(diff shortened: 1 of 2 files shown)


schemas/aws/policy-1.yml
```diff
@@ -18,6 +18,8 @@ properties:
     type: string
   description:
     type: string
+  mandatory:
+    type: boolean
```
geoberle, line 22: please add a `description`

maorfr: added in https://github.com/app-sre/qontract-schemas/pull/81/commits/e09e71dde11d20d358e759e5562a440e42638c84

geoberle (approved):
LGTM

PR #83, opened 2022-02-21
creating new schema for tf state

suzana-nesic:
Part of [APPSRE-4516](https://issues.redhat.com/browse/APPSRE-4516)

Goal is to have a separate yaml file within app-interface that is then referenced in the account.yml file:
```
---
$schema: /dependencies/terraform-state-1.yml

provider: s3
bucket: someBucket
region: someRegion
integrations:
- integration: terraform-resources
  key:  qontract-reconcile.tfstate
- integration: terraform-tgw-attachements
  key: qontract-reconcile-tgw-attachments.tfstate
```
Within some-cloud-account.yml:
```
...
terraformState:
  $ref: /aws/app-sre-stage/tf-state-test.yml
```

For app-sre, we have specific terraform integrations for route53, resources, tgw, users, and vpc peerings

Signed-off-by: Suzana Nesic <snesic@redhat.com>

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -807,7 +807,7 @@
   - { name: premiumSupport, type: boolean, isRequired: true }
   - { name: partition, type: string }
   - { name: sharing, type: AWSAccountSharingOption_v1, isList: true, isInterface: true }
-  - { name: terraformState, type: AWSTerraformState_v1, isList: true, isRequired: false, isInterface: true }
+  - { name: terraformState, type: AWSTerraformState_v1, isRequired: false, isInterface: true }
   - name: ecrs
     type: AWSECR_v1
     isList: true
@@ -919,17 +919,22 @@
     strategy: fieldMap
     field: provider
     fieldMap:
-      aws-bucket: AWSBucketInfo_v1
+      s3: AWSS3TerraformState_v1
   fields:
-  - { name: provider, type: string, isRequired: true }
+  - { name: provider, type: string, isRequired: false }
 
-- name: AWSBucketInfo_v1
+- name: AWSS3TerraformState_v1
   interface: AWSTerraformState_v1
   fields:
+  - { name: provider, type: string, isRequired: true}
   - { name: bucket, type: string, isRequired: true }
-  - { name: key, type: string, isRequired: true }
   - { name: region, type: string, isRequired: true }
+  - { name: integrations, type: AWSTerraformStateIntegrations_v1, isList: true, isRequired: true }
+
+- name: AWSTerraformStateIntegrations_v1
+  fields:
   - { name: integration, type: string, isRequired: true }
+  - { name: key, type: string, isRequired: true }
 
 - name: ACMDomain_v1
   fields:
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -915,13 +915,10 @@
 
 - name: AWSTerraformState_v1
   fields:
-  - { name: terraform_resources, type: AWSTerraformResources_v1, isRequired: true }
-
-- name: AWSTerraformResources_v1
-  fields:
   - { name: bucket, type: string, isRequired: true }
   - { name: key, type: string, isRequired: true }
   - { name: region, type: string, isRequired: true }
+  - { name: integration, type: string, isRequired: true }
 
 - name: ACMDomain_v1
   fields:
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -807,7 +807,7 @@
   - { name: premiumSupport, type: boolean, isRequired: true }
   - { name: partition, type: string }
   - { name: sharing, type: AWSAccountSharingOption_v1, isList: true, isInterface: true }
-  - { name: terraformState, type: AWSTerraformState_v1, isRequired: false, isInterface: true }
+  - { name: terraformState, type: TerraformState_v1, isRequired: false, isInterface: true }
   - name: ecrs
     type: AWSECR_v1
     isList: true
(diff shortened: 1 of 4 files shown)


schemas/aws/terraform-state-1.yml
```diff
@@ -0,0 +1,17 @@
...
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /aws/account-1.yml
```
geoberle, line 11: this line needs to mention the schema itself. have a look at how other schema files are doing this

geoberle (commented):
thanks suzana. please also update the graphql-schemas/schema.yml file to reflect the change. you will also need to make a relation from the account to this schema here.

schemas/aws/terraform-state-1.yml
suzana-nesic: Done, thank you for catching that!

suzana-nesic:
I believe I have it referenced in the schema file in the simplest way possible? Or is there an example that does what I want in an even shorter step?

schemas/aws/account-1.yml
```diff
@@ -103,6 +103,12 @@ properties:
     enum:
     - aws
     - aws-us-gov
+  terraformState:
+    type: array
+    description: key information for a terraform's state location and integration
+    terraform_resources:
```
geoberle, line 109:
```suggestion
    items:
```

geoberle: this needs to be named `items` following the openapi spec

graphql-schemas/schema.yml
```diff
@@ -709,6 +709,7 @@
   - { name: resetPasswords, type: AWSAccountResetPassword_v1, isList: true }
   - { name: premiumSupport, type: boolean, isRequired: true }
   - { name: partition, type: string }
+  - { name: terraformState, type: AWSTerraformState_v1, isList: true }
```
geoberle, line 712: with everything that is going on in https://github.com/app-sre/qontract-schemas/pull/87 i wonder if we should also support other terraform state backends and reference an interface type instead of a concrete one like [here](https://github.com/app-sre/qontract-schemas/pull/87/files#diff-6bad3cee90319b7494304e3ce4e439992becd03f9d8825eca7f44687f8184ee9R1351) 

cc @maorfr

rporres: That is a great idea.

maorfr: makes sense to me!

schemas/aws/account-1.yml
```diff
@@ -103,6 +103,12 @@ properties:
     enum:
     - aws
     - aws-us-gov
+  terraformState:
+    type: array
```
maorfr, line 107: why is this an array?

schemas/aws/terraform-state-1.yml
```diff
@@ -0,0 +1,20 @@
...
+    type: string
+    enum:
+    - /aws/terraform-state-1.yml
+  bucket:
+    type: string
+  key:
+    type: string
+  region:
+    type: string
+  integration:
```
maorfr, line 19: what is the `integration` field for? mind adding descriptions?

schemas/aws/account-1.yml
suzana-nesic: We want to be able to reference the terraform resource schema multiple times because there will be multiple integrations for the same bucket

suzana-nesic:
@geoberle @maorfr I can clean up the last commit I did, but is that all that is needed to implement the interface type?

schemas/aws/terraform-state-1.yml
geoberle: so each /dependencies/terraform-state-1.yml object is a bucket/key reference for a specific integration and a list of those can be referenced from an /aws/account-1.yml, right?

an alternative would have been to reference terraform-state-1 objects as a dictionary from an account, having the key an integration name and the value a terraform-state-1 object. but this would imply updating the account schema every time we have a new integration.

geoberle:
> @geoberle @maorfr I can clean up the last commit I did, but is that all that is needed to implement the interface type?
yes i think there was a misunderstanding. the last commit made the AWSAccount an interface. instead the the object representing the terraform state would make a good interface and the placement into a bucket an implementation of that interface. does that make sense?

suzana-nesic:
> > @geoberle @maorfr I can clean up the last commit I did, but is that all that is needed to implement the interface type?
> > yes i think there was a misunderstanding. the last commit made the AWSAccount an interface. instead the the object representing the terraform state would make a good interface and the placement into a bucket an implementation of that interface. does that make sense?

I think I'm following. I'll go ahead and make another commit and see if what I think I understand is correct.

suzana-nesic:
@geoberle I think the team is happy with adding an interface and keeping it AWS-specific for now. There is a ticket linked in my design doc MR from Maor to make it more open for different cloud providers that can be revisited another time. Whenever convenient, could you take a look at the latest changes and let me know if this was done correctly?

graphql-schemas/schema.yml
```diff
@@ -912,6 +913,29 @@
   - { name: filter_prefix, type: string}
   - { name: filter_suffix, type: string}
 
+- name: AWSTerraformState_v1
```
maorfr, line 916:
```suggestion
- name: TerraformState_v1
```

graphql-schemas/schema.yml
```diff
@@ -807,6 +807,7 @@
   - { name: premiumSupport, type: boolean, isRequired: true }
   - { name: partition, type: string }
   - { name: sharing, type: AWSAccountSharingOption_v1, isList: true, isInterface: true }
+  - { name: terraformState, type: TerraformState_v1, isRequired: false, isInterface: true }
```
maorfr, line 810:
```suggestion
  - { name: terraformState, type: TerraformStateAWS_v1, isRequired: false }
```

schemas/dependencies/terraform-state-1.yml
```diff
@@ -0,0 +1,31 @@
...
+    type: array
+    description: terraform state integration file
+    items: 
+      type: object
+    properties:
+      integration:
+        type: string
+      key:
+        type: string
+    additionalProperties: false
```
maorfr, line 31: missing newline

maorfr, line 31: this section should be indented 2 to the right

schemas/dependencies/terraform-state-1.yml
```diff
@@ -0,0 +1,31 @@
...
+    type: string
+    enum:
+    - s3
+  bucket:
+    type: string
+  region:
+    type: string
+  integrations:
+    type: array
+    description: terraform state integration file
```
geoberle, line 23:
```suggestion
    description: holds information which bucket key stores the terraform state per integration
```

geoberle (approved):
LGTM

maorfr:
promoting and adding example usage in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/40034

PR #85, opened 2022-02-22
route53-zone add name field

maorfr:
following https://github.com/app-sre/qontract-reconcile/pull/1962 where we added support for dns zones, we need to be able to provision resources with names that includes dots.

### DIFF
--- a/schemas/openshift/terraform-resource-1.yml
+++ b/schemas/openshift/terraform-resource-1.yml
@@ -16,6 +16,8 @@ properties:
   account:
     "$ref": "/aws/tenant_accounts-1.yml#/properties/account"
   identifier:
+    "$ref": "/common-1.json#/definitions/longIdentifier"
+  name:
     type: string
   variables:
     type: object
@@ -813,6 +815,8 @@ oneOf:
       "$ref": "/aws/tenant_accounts-1.yml#/properties/account"
     identifier:
       type: string
+    name:
+      type: string
     region:
       "$ref": "/aws/regions-1.yml#/properties/region"
     output_resource_name:
@@ -822,5 +826,6 @@ oneOf:
   required:
   - account
   - identifier
+  - name
 required:
 - provider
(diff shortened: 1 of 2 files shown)


maorfr:
required for https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34050

rporres:
I don't know if "some identifier sanity checks in qontract-reconciler" is enough and id there are tests that will fail if those are removed. My general view is that we must always do the schema as tight as possible and only add checks in q-r for those things that cannot be done via schema.

The only way I can think of making this tighter is via `oneOf` via the provider. wdyt?

rporres:
what's the difference between `name` and `identifier`=

maorfr:
> Is that something that you will add to the integration

indeed: https://github.com/app-sre/qontract-reconcile/pull/2195

rporres:
understood. What's `name` exactly? The base `fqdn` of the zone?

maorfr:
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/route53_zone#name

indeed

rporres:
So then I understand that it needs validation too, isn't it? And a `description` explaining what it is in the context of the `route53` provider... :)

maorfr:
added description here

schemas/openshift/terraform-resource-1.yml
```diff
@@ -813,6 +815,10 @@ oneOf:
       "$ref": "/aws/tenant_accounts-1.yml#/properties/account"
     identifier:
       type: string
+    name:
+      description: base fqdn of the zone
+      type: string
+      format: uri
```
rporres, line 821: no, uris are more than we need

This should be better https://github.com/app-sre/qontract-schemas/blob/main/schemas/common-1.json#L47-L51 since k8s container names are valid DNS names

maorfr: done in https://github.com/app-sre/qontract-schemas/pull/85/commits/1569085eddfa124dfa752b37818ebf4760244dd4

rporres (approved):

PR #87, opened 2022-02-28
[terraform-resources] introduce managedProviderResources

maorfr:
replaces #86

this PR adds support for a new way to define terraform resources, which is future compatible with adding additional providers, except for AWS. this is backwards compatible with the option to define AWS resources only under `terraformResources`.

the new section looks like this:
```
terraformProviderResources:
- provider: aws
  provisioner:
    $ref: /aws/app-int-example-01/account.yml
  resources:
  - provider: rds
    identifier: app-int-example-01-rds
    defaults: /terraform/rds/free-tier-1.yml
    output_resource_name: creds
```

this approach follows the provider pattern as described here: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-interface/qontract-reconcile-patterns.md#the-provider-pattern
> In this pattern, we allow defining items in a similar way while handling them differently behind the scenes.
> Each item is handled a bit differently in our integrations, while allowing users to define all resources in a similar way or even in a single location.


patjlm:
+1 Nice

patjlm (approved):
LGTM

graphql-schemas/schema.yml
```diff
@@ -854,7 +864,55 @@
   - { name: vault_tls_secret_path, type: string }
   - { name: vault_tls_secret_version, type: int }
 
-- name: NamespaceTerraformResource_v1
+- name: NamespaceTerraformProviderResource_v1
```
geoberle, line 867: the name suggests that within this schema, one resource would be found, instead it contains all resources belonging to a certain provisioner. this is a change

graphql-schemas/schema.yml
```diff
@@ -879,7 +937,7 @@
       alb: NamespaceTerraformResourceALB_v1
       secrets-manager: NamespaceTerraformResourceSecretsManager_v1
       asg: NamespaceTerraformResourceASG_v1
-      route53-zone: NamespaceTerraformResourceRoute53Zone_v1
+      dns-zone: NamespaceTerraformResourceRoute53Zone_v1
   fields:
   - { name: provider, type: string, isRequired: true }
   - { name: account, type: string, isRequired: true }
```
geoberle, line 943: why is the account still mandatory if it is practically the provisioner one level up?

graphql-schemas/schema.yml
maorfr: resolved by renaming to NamespaceTerraformProvisionProvider_v1

graphql-schemas/schema.yml
maorfr: added a comment about backwards compatibility

geoberle (approved):
LGTM

lechuk47:
Taking into account the providers we are going to use work quite different, I think it's better to maintain the providers just at the resource level and not group them by a cloud provider.  
e,g: 

```yaml
resources:
- provider: aws-route53
  provisioner:
   $ref: /aws/account.yml 
  attr: ...
  attr: ...

- provider: gcp-cloud-dns 
  provisioner: 
    $ref: /gcp/myorg/gcp-project-1.yml 
  attr: ...
  attr: ...

- provider: gcp-iam-group
  provisioner:
    $ref: /gcp/myorg/folder.yml
  attr: ... 
  attr: ...
```

GCP and AWS have completely separated terraform resource names and the features, way to configure the components differs a lot. I think it's better to just have different providers at the resource level and not share them in a cloud provider spec. This way we have all the freedom to set just the real and required attributes by each component and not have mixed schemas and logic for different cloud providers. 

Additional notes: 
* In GCP, not all resources are bound to a project. The provisioner could have a reference to the account where it belongs. 
* Maybe the DNS is the service that have more similarities, but If we consider the Network, compute, compute managed groups, etc. The differences are much more accentuated. 
* Maybe it implies more code, but the maintainability and safety to change code of a single resource it's worth the effort IMO.

maorfr:
i added a design document in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34377

just to answer some things here:
1. grouping is an implementation detail IMO. I think it's good because in many namespace files we have many resources provisioned in the same account, so this approach will reduce some duplication. how is this grouping changing our ability to "set just the real and required attributes by each component and not have mixed schemas and logic for different cloud providers"?
2. about gcp projects vs accounts, indeed, this is one of the things this approach solves. we can have a provider called `gcp-project` and one called `gcp-account`.
3. not sure i understand how the last 2 comments are related.

lechuk47:
OK, after reading the design doc it's more clear to me. I was concerned about the `dns-zone` on your example. I thought you were suggesting to share providers across both cloud providers. Reading the doc it seems that it's not the case and it's just a grouping change.

maorfr:
indeed!
i do plan to later "expose" both route53 zone and gcp managed zone as `dns-zone`, but due to the connection to the provisioner, the implementation will be completely different. it's about abstraction, and not about reusing code or anything.

lechuk47 (approved):
lgtm

rporres:
Now that the design doc has been approved and the interface has settled it is a god moment to say this. I know asking this is trying to fix a schema documentation debt of almost three years in some cases, but some documentation at least in the new fields wouldn't hurt at all 😬 

LGTM otherwise

PR #88, opened 2022-02-28
APPSRE-3341: Fix owned_saas_files json schema definition

lechuk47:
Fix schema for `owned_saas_files` in `role-1` definition

### DIFF
--- a/schemas/access/role-1.yml
+++ b/schemas/access/role-1.yml
@@ -101,11 +101,15 @@ properties:
   owned_saas_files:
     type: array
     items:
-      anyOf:
-      - "$ref": "/common-1.json#/definitions/crossref"
-        "$schemaRef": "/app-sre/saas-file-1.yml"
-      - "$ref": "/common-1.json#/definitions/crossref"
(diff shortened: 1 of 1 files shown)


maorfr (approved):
AMAZING

schemas/access/role-1.yml
```diff
@@ -101,11 +101,15 @@ properties:
...
-        "$schemaRef": "/app-sre/saas-file-2.yml"
+      "$ref": "/common-1.json#/definitions/crossref"
+      "$schemaRef":
+        type: object
+        properties:
+          '$schema':
+            type: string
+            enum:
+              - "/app-sre/saas-file-1.yml"
+              - "/app-sre/saas-file-2.yml"
```
rporres, line 112: nit: The general style of the document is to avoid indenting arrays.

rporres (approved):
Nice solution.

Just to add steps into solving our general lack of description in the schema, if you could just add a small `description` into this field you're fixing, it would be great.

schemas/access/role-1.yml
lechuk47: changed.

PR #89, opened 2022-02-28
Add docs on crossref usage/definition

lechuk47:
- Add docs on crossref definitions


geoberle (approved):
LGTM - thank you for writing this

PR #90, opened 2022-02-28
APPSRE-4403: Update app-interface settings schema for LDAP

bhushanthakur93:


graphql-schemas/schema.yml
```diff
@@ -18,12 +18,18 @@
   - { name: pushGatewayCluster, type: Cluster_v1 }
   - { name: alertingServices, type: string, isList: true }
   - { name: endpointMonitoringBlackboxExporterModules, type: string, isList: true }
+  - { name: ldap, type: LdapSettings_v1, isRequired: true}
```
maorfr, line 21: for the sake of being "open source", i suggest we don't make this required

maorfr (approved):
otherwise LGTM

graphql-schemas/schema.yml
bhushanthakur93: Yes, that is a good point. I'll make the change.

PR #91, opened 2022-03-02
add aws sharing options

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4621

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/34669

this PR adds support to define `sharing` between aws accounts. we use the provider pattern to allow future sharing options between aws accounts.

this `sharing` option is defined on the source account.

example:
```
$schema: /aws/account-1.yml

name: source
...
sharing:
- provider: ami
  account:
    $ref: /aws/destination/account.yml
  regex: '^.*$'
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -774,6 +774,7 @@
   - { name: provider, type: string, isRequired: true }
   - { name: account, type: AWSAccount_v1, isRequired: true }
   - { name: regex, type: string, isRequired: true }
+  - { name: region, type: string }
 
 - name: AWSUserPolicy_v1
   fields:
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -716,6 +716,7 @@
   - { name: resetPasswords, type: AWSAccountResetPassword_v1, isList: true }
   - { name: premiumSupport, type: boolean, isRequired: true }
   - { name: partition, type: string }
+  - { name: sharing, type: AWSAccountSharingOption_v1, isList: true, isInterface: true }
   - name: ecrs
     type: AWSECR_v1
     isList: true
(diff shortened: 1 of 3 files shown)


graphql-schemas/schema.yml
```diff
@@ -756,6 +757,24 @@
   - { name: user, type: User_v1, isRequired: true }
   - { name: requestId, type: string, isRequired: true }
 
+- name: AWSAccountSharingOption_v1
+  isInterface: true
```
geoberle, line 761: what is the idea here? e.g. also share stuff like RDS snapshots etc?

graphql-schemas/schema.yml
```diff
@@ -756,6 +757,24 @@
...
+  fields:
+  - { name: provider, type: string, isRequired: true }
+  - { name: account, type: AWSAccount_v1, isRequired: true }
+
+- name: AWSAccountSharingOptionAMI_v1
+  interface: AWSAccountSharingOption_v1
+  fields:
+  - { name: provider, type: string, isRequired: true }
+  - { name: account, type: AWSAccount_v1, isRequired: true }
+  - { name: regex, type: string, isRequired: true }
```
geoberle, line 776: please add a description. is this a regex of AMI IDs?

graphql-schemas/schema.yml
maorfr: perhaps. it's a future compatible design.

graphql-schemas/schema.yml
maorfr: there is a description in the json schema?

schemas/aws/sharing-option-1.yml
```diff
@@ -0,0 +1,43 @@
...
+    type: string
+    description: type of sharing to implement
+    enum:
+    - ami
+  account:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef": "/aws/account-1.yml"
+  regex:
+    type: string
+    description: regex expression to filter items by
```
geoberle, line 23: ... by what? :D

graphql-schemas/schema.yml
geoberle: ok see comment below

schemas/aws/sharing-option-1.yml
maorfr: The items which we are in the context of?

geoberle: i just realized it is an optional field. my concern was that regex would be not flexible enough to be the way of filtering for all sharing options to come (e.g. filtering by multiple tags)

geoberle (approved):
Thanks. This change is great and has potential for a lot of interesting features. LGTM

PR #92, opened 2022-03-10
add assume_condition

BumbleFeng:
[APPSRE-4674](https://issues.redhat.com/browse/APPSRE-4674)

add condition context keys for IAM and AWS STS https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_iam-condition-keys.html

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1052,6 +1052,7 @@
   - { name: account, type: string, isRequired: true }
   - { name: identifier, type: string, isRequired: true }
   - { name: assume_role, type: AssumeRole_v1, isRequired: true }
+  - { name: assume_condition, type: json }
   - { name: inline_policy, type: json }
   - { name: output_resource_name, type: string }
   - { name: annotations, type: json }
--- a/schemas/openshift/terraform-resource-1.yml
+++ b/schemas/openshift/terraform-resource-1.yml
@@ -33,6 +33,8 @@ properties:
     type: object
   assume_role:
     type: object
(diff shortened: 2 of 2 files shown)


rporres:
Can you provide some context for this change?

rporres:
Are these the kind of conditions that we can add here? https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html

If so, it would be nice for the schema reader to add a small description about this field, what it exactly means and possibly that very link 😄 

Or maybe https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_iam-condition-keys.htm . It has taken me a bit to understand what was all of this 😅

BumbleFeng:
> Are these the kind of conditions that we can add here? https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html
> 
> If so, it would be nice for the schema reader to add a small description about this field, what it exactly means and possibly that very link 😄
> 
> Or maybe https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_iam-condition-keys.html. It has taken me a bit to understand what was all of this 😅

Updated. This is for [condition context keys for IAM and AWS STS](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_iam-condition-keys.html)  with [format](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition.html)

rporres (approved):

PR #95, opened 2022-03-14
[terraform-resources] ASG add support for upstream job

maorfr:
part of https://issues.redhat.com/browse/APPSRE-3925

now that we have support for continuous deployment of AMIs, we want to avoid failing under healthy conditions.
for example, once a new commit is pushed, the integration tries to use a corresponding AMI. initially, that AMI is being built, which means it doesn't exist yet.

to be able to not fail in this condition, and to improve visibility from the ASG definition, this PR adds support for an `upstream` job to wait for (like saas files).

### DIFF
--- a/schemas/openshift/terraform-resource-1.yml
+++ b/schemas/openshift/terraform-resource-1.yml
@@ -139,6 +139,13 @@ properties:
       type: string
   image:
     type: object
+    properties:
+      upstream:
+        type: object
+        properties:
+          instance:
+            "$ref": "/common-1.json#/definitions/crossref"
+            "$schemaRef": "/dependencies/jenkins-instance-1.yml"
 oneOf:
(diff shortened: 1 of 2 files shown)


BumbleFeng (approved):

schemas/openshift/terraform-resource-1.yml
```diff
@@ -139,6 +139,13 @@ properties:
       type: string
   image:
     type: object
+    properties:
+      upstream:
+        type: object
+        properties:
```
geoberle, line 145: isn't `name` missing up here? i'm unsure if it is really necessary though. i was not able to find respective openapi docs. up until now i only dealt with examples where the config in the top level `properties` section matches the ones in the `oneOf` section

geoberle (approved):

PR #97, opened 2022-03-15
APPSRE-4404 schema for githubRepoInvites

fishi0x01:
**Motivation**

We currently store configuration for the `github-repo-invites` integration in our `config.toml`. This PR adds a schema to move the `[github-repo-invites]` toml section to our `app-interface` repo.

**Test**

Successfully ran:

```
make bundle
```

Further, tested with `--dry-run` in corresponding [q-r PR](https://github.com/app-sre/qontract-reconcile/pull/2239) in qontract-reconcile

### DIFF
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -52,12 +52,14 @@ properties:
     type: object
     additionalProperties: false
     properties:
-      description:
-        type: string
       credentials:
         "$ref": "/common-1.json#/definitions/vaultSecret"
+    description: |
+      Configuration for github-repo-invites integration.
+      Currently contains information for fetching the
+      token of the github app-sre bot for accepting
+      invitations to repositories.
     required:
-    - description
     - credentials
   ldap:
     type: object
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -48,6 +48,12 @@ properties:
         type: string
       credentials:
         "$ref": "/common-1.json#/definitions/vaultSecret"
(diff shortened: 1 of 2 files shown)


schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -48,6 +48,12 @@ properties:
         type: string
       credentials:
         "$ref": "/common-1.json#/definitions/vaultSecret"
+  githubRepoInvites:
+    type: object
```
rporres, line 52: Please add descriptions for the new fields.

schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -48,6 +48,12 @@ properties:
...
+    additionalProperties: false
+    properties:
+      credentials:
+        "$ref": "/common-1.json#/definitions/vaultSecret"
```
rporres, line 56: I guess you want to make `credentials` required, right?

rporres (commented):
app-interface settings is becoming a mixed bag of too many loosely related things, but I don't think we have another place for this kind of information

schemas/app-interface/app-interface-settings-1.yml
fishi0x01: Fixed

rporres: No, not that way 😄  A description field that documents the new field, e.g. https://github.com/app-sre/qontract-schemas/blob/main/schemas/app-sre/slo-document-1.yml#L42

fishi0x01: I see. Adjusted it now :)

rporres (approved):

PR #103, opened 2022-03-21
reduce number of required attributes

dwelch0:
Reduce amount of attributes required to configure for vault oidc roles. 

Ticket: https://issues.redhat.com/browse/APPSRE-4659

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -215,17 +215,17 @@
   fields:
   - { name: _type, type: string, isRequired: true }
   - { name: allowed_redirect_uris, type: string, isList: true, isRequired: true }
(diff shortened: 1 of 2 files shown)


TGPSKI (approved):

maorfr:
while removing required fields is probably ok for oidc, are they still needed for other methods?

i think the provider pattern may fit in here? https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-interface/qontract-reconcile-patterns.md#the-provider-pattern

dwelch0:
The fields that I removed `required` on are unique to oidc roles. In fact, I initially attempted to remove it from other fields that are shared between both roles and received an error that validator couldn't resolve required version of a field from unrequired for same field (ex: token_ttl).

PR #106, opened 2022-03-23
implement schema to support vault entities and aliases

dwelch0:
TICKETS: 
* https://issues.redhat.com/browse/APPSRE-4656
* https://issues.redhat.com/browse/APPSRE-4657


graphql-schemas/schema.yml
```diff
@@ -1957,6 +1957,21 @@
...
+  fields:
+  - { name: name, type: string, isRequired: true }
+  - { name: auth_type, type: string, isRequired: true }
+  - { name: custom_metadata, type: json }
+
+- name: VaultEntity_v1
+  fields:
+  - { name: name, type: string, isRequired: true }
+  - { name: type, type: string, isRequired: true }
+  - { name: policies, type: string, isList: true, isRequired: true }
```
geoberle, line 1970: could this be a list of references to `VaultPolicy_v1` objects as well?

dwelch0: I briefly considered this early on but didn't pursue it because I had no need for the rules portion of ```VaultPolicy_v1```. But I ended up needing to write a policy validation check within the logic to ensure listed policies existed.. This wouldn't be necessary if I had used ```VaultPolicy_v1```. Thanks for bringing up, I am going to go back and update the schema/rewrite the logic to handle these objects instead.

PR #118, opened 2022-03-31
Added contract related schemas

rporres:
part of [APPSRE-4788](https://issues.redhat.com/browse/APPSRE-4788)

This is an implementation of the accepted proposal from @maorfr in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/32536.

It adds the link to the SRE checkpoint schema. This is optional for the moment, I will make it mandatory once I have updated all the checkpoint files.

### DIFF
--- a/schemas/app-sre/contract-1.yml
+++ b/schemas/app-sre/contract-1.yml
@@ -1,0 +1,34 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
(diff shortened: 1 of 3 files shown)


maorfr:
I think my proposal was to add a link to a contract version from an app file?
This makes sense as well, so perhaps both?

rporres:
Wouldn't that be redundant? In order to get the last checkpoint we can just use the `app` reference and sort by date.

maorfr (approved):
LGTM as is

we can add a contract version to an app later as well

PR #119, opened 2022-04-02
generic-secret output_format for terraform based output secrets

geoberle:
### What changes?
Provide an output_format section to /openshift/terraform-resource-1.yml. This section will hold configuration data to drive the formatting process for terraform output data. output_format configuration is provider based, allowing different implementations to format the terraform outputs into consumable kubernetes secrets.

The first supported provider is the generic-secret provider, which formats the terraform output data as a kubernetes secret in the exact same way as it was handled before this change. to influence the output format, the secret data section can be specified inline, while providing jinja2 templating functionality.

### Example
The generic-secret provider will be the implicit default one when none is specified, making the following three declarations equivalent:

```yaml
  terraformResources:
  - provider: aws-iam-service-account
    ...
```

```yaml
  terraformResources:
  - provider: aws-iam-service-account
    output_format:
      provider: generic-secret
    ...
```

```yaml
  terraformResources:
  - provider: aws-iam-service-account
    output_format:
      provider: generic-secret
      data:
        aws_access_key_id: {{ aws_access_key_id }}
        aws_secret_access_key: {{ aws_secret_access_key }}
    ...
```

Another example shows how AWS access credentials can be formatted in the credentials file format

```yaml
  terraformResources:
  - provider: aws-iam-service-account
    output_format:
      provider: generic-secret
      data:
        credentials: |
          [default]
          aws_access_key_id = {{ aws_access_key_id }}
          aws_secret_access_key = {{ aws_secret_access_key }}

    ...
```

### References
Part of https://issues.redhat.com/browse/APPSRE-4766

### DIFF
--- a/schemas/openshift/terraform-output-format-1.yml
+++ b/schemas/openshift/terraform-output-format-1.yml
@@ -1,0 +1,43 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /openshift/terraform-output-format-1.yml
+  provider:
+    type: string
+    enum:
+    - generic-secret
+  data:
+    type: object
+    additionalProperties: false
+    patternProperties: {
+      "^[a-zA-Z0-9_]{0,253}$": {
+        "type": "string"
+      }
+    }
+oneOf:
+- additionalProperties: false
+  properties:
+    provider:
+      type: string
+      enum:
+      - generic-secret
+    data:
+      type: object
+      additionalProperties: false
+      patternProperties: {
+        "^[a-zA-Z0-9_]{0,253}$": {
+          "type": "string"
+        }
+      }
+      description: |
+        the data to use for the secret. the field values can use the
+        terraform output variables as jinja2 placeholders
+  required:
+  - provider
--- a/schemas/openshift/terraform-output-format-1.yml
+++ b/schemas/openshift/terraform-output-format-1.yml
@@ -11,8 +11,10 @@ properties:
     - /openshift/terraform-output-format-1.yml
   provider:
     type: string
-    enum:
-    - generic-secret
+    description: |
+      the provider defines the method of formatting the terraform output
+      into a secret. a provider can declare additional configuration data to
+      get the job done or define the format directly
   data:
     type: object
     additionalProperties: false
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -996,7 +996,24 @@
   - { name: account, type: string, isRequired: true }
   - { name: identifier, type: string, isRequired: true }
   - { name: output_resource_name, type: string }
+  - { name: output_format, type: NamespaceTerraformResourceOutputFormat_v1, isInterface: true }
   - { name: annotations, type: json }
+
+- name: NamespaceTerraformResourceOutputFormat_v1
+  isInterface: true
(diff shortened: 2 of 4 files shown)


schemas/openshift/terraform-output-format-1.yml
```diff
@@ -0,0 +1,43 @@
...
+    type: string
+    enum:
+    - generic-secret
+  data:
+    type: object
+    additionalProperties: false
+    patternProperties: {
+      "^[a-zA-Z0-9_]{0,253}$": {
+        "type": "string"
+      }
```
rporres, line 22: ❤️

schemas/openshift/terraform-output-format-1.yml
```diff
@@ -0,0 +1,43 @@
...
+type: object
+
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /openshift/terraform-output-format-1.yml
+  provider:
+    type: string
```
rporres, line 13: Can you add a `description` that clarifies that this is actually the format? I understand the use of the `provider` name so that it is understood that we're using the provider pattern.

schemas/openshift/terraform-output-format-1.yml
```diff
@@ -0,0 +1,43 @@
...
+    - generic-secret
+  data:
+    type: object
+    additionalProperties: false
+    patternProperties: {
+      "^[a-zA-Z0-9_]{0,253}$": {
+        "type": "string"
+      }
+    }
+oneOf:
```
rporres, line 24: Will this be expanded in the future with the well-known formats mentioned in https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/terraform-output-secret-format.md#well-known-formats?

schemas/openshift/terraform-output-format-1.yml
```diff
@@ -0,0 +1,43 @@
...
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /openshift/terraform-output-format-1.yml
+  provider:
+    type: string
+    enum:
+    - generic-secret
```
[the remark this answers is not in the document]
geoberle: ah you are right. it is not required "up there"

schemas/openshift/terraform-output-format-1.yml
geoberle: yes, well known formats could be dedicated providers or also something like
```yaml
terraformResources:
  - provider: aws-iam-service-account
    output_format:
      provider: well-known-formats
      format: aws-credentials-file
    ...
```

schemas/openshift/terraform-output-format-1.yml
geoberle: [d0c3733](https://github.com/app-sre/qontract-schemas/pull/119/commits/d0c37333c1a7411e7156a4d0c9c39aa2d5f8b6ab)

schemas/openshift/terraform-output-format-1.yml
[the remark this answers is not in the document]
geoberle: [d0c3733](https://github.com/app-sre/qontract-schemas/pull/119/commits/d0c37333c1a7411e7156a4d0c9c39aa2d5f8b6ab)

schemas/openshift/terraform-output-format-1.yml
rporres: great, thx

PR #120, opened 2022-04-08
allow saas targets to reference repo branches prefixed with release-

geoberle:
hypershift maintains release branches and the release- prefix is a well established pattern for projects within the github openshift organization

ref: https://issues.redhat.com/browse/APPSRE-4832

### DIFF
--- a/schemas/app-sre/saas-file-2.yml
+++ b/schemas/app-sre/saas-file-2.yml
@@ -103,7 +103,7 @@ properties:
               ref:
                 type: string
                 # pattern: '^([0-9a-f]{40}|master|main|quayio|development-4\.0|internal|mk-release)$'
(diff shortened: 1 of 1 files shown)


patjlm (approved):
LGTM

maorfr:
Can we discuss this next week?

maorfr:
the reason we decided not to go with this direction is -
release branches indicate "production". using a release branch as the ref means that there is continuous deployment to production, and this is not a pattern we encourage.

it also means that the deployment to production will need to be linked to an "upstream" job that builds the release branch, and that means that images deployed to production may not be tested in stage first.

further reading on the selected approach: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/the-app-sre-hotfix-path.md

PR #122, opened 2022-04-13
APPSRE-3556 deprecate saas-file-1

fishi0x01:
All saas files are moved to tekton. We can deprecate `saas-file-1` now.


janboll:
:question: Still found a couple of references in app-interface

rporres (approved):
To be fair, you're not deprecating, you're actually removing the support for saas-file-1.yml files :-D

LGTM otherwise

fishi0x01:
@janboll good catch! Thank you! :bow: 
Removed the occurrences in app-interface.

maorfr:
just like we add schemas before logic, we should remove logic before schemas.

going to revert this one to promote the next PR, but this by no means mean this PR is not good :)

PR #139, opened 2022-05-06
Support multiple vault instances

dwelch0:
This PR updates Vault config schemas in effort for vault manager to support reconciling multiple vault instances.

**IMPORTANT:** This PR will make `instance` a required attribute on existing vault resource definitions. Ensure that MR is made to App Interface for the following before promoting qontract-schema image:
1. Add instance definition
2. Add instance attribute reference to affected schemas

**Ticket:** https://issues.redhat.com/browse/APPSRE-4506


schemas/vault-config/instance-1.yml
```diff
@@ -0,0 +1,54 @@
...
+  authType:
+    type: string
+  address:
+    type: string
+  roleID:
+    "$ref": "/common-1.json#/definitions/vaultSecret"
+  secretID:
+    "$ref": "/common-1.json#/definitions/vaultSecret"
+  token:
+    "$ref": "/common-1.json#/definitions/vaultSecret"
```
maorfr, line 27: since all these will likely be in the same secret, do we want to define a single field?

graphql-schemas/schema.yml
```diff
@@ -1857,6 +1857,45 @@
...
+    field: authType
+    fieldMap:
+      approle: VaultInstanceApprole_v1
+      token: VaultInstanceToken_v1
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: labels, type: json }
+  - { name: name, type: string, isRequired: true }
+  - { name: description, type: string, isRequired: true }
+  - { name: authType, type: string, isRequired: true }
```
maorfr, line 1889: i'd recommend to create an `auth` section which uses the provider pattern: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-interface/qontract-reconcile-patterns.md#the-provider-pattern

that way, a vault instance doesn't need to be an interface. it's really only the auth, isn't it?

schemas/access/oidc-permission-1.yml
```diff
@@ -28,6 +28,15 @@ properties:
...
             - /vault-config/policy-1.yml
+  instance:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef":
+      type: object
+      properties:
+        '$schema':
+          type: string
+          enum:
+          - /vault-config/instance-1.yml
```
maorfr, line 39: what's wrong with:
```
"$ref": "/common-1.json#/definitions/crossref"
"$schemaRef": "/vault-config/instance-1.yml"
```

graphql-schemas/schema.yml
dwelch0: Good point. This approach would be cleaner. I will work on implementing

schemas/vault-config/instance-1.yml
dwelch0: I agree that it is redundant to specify secret path multiple times across different `VaultSecrets`, since it is highly likely the fields will be within the same secret. However, I believe explicitly defining the `field` attribute within each `VaultSecret` is less error prone/management.

The `ReadSecret` function pulls the full json for a particular secret path and `field` is utilized to quickly reference a specific value. Without defining fields, we would need to enforce naming conventions for each attribute when originally creating the vault secret (ex: role id should be entered as `role-id` etc.) and vault-manager logic would need to be hard-coded to specifically look for `role-id` or whatever we define.

By allowing users to define the name to map to roleID etc., we don't need to bother with enforcing naming. `roleID` can map to whatever field name they defined. This is demonstrated in the refactored test suite in which one instance defines token field as name `token` and the other defines name to look for as `rootToken`.

schemas/access/oidc-permission-1.yml
dwelch0: Not a thing. Copied a different ref example thinking it was a standard approach. Changed references to simpler form

graphql-schemas/schema.yml
dwelch0: Implemented

maorfr (approved):
LGTM

very nice!

PR #141, opened 2022-05-16
add app as mandatory field to slo document

maorfr:
this will allow to find out in a simple way if an app has an slo document.


rporres (commented):
yeah, this will make sense for apps that share a namespace. 

Since it is a breaking change, why don't you split it in two:

- Add app to the slo doc schema
- Make it mandatory

So you won't block any potential schema change while you're updating all the data in app-interface and getting it reviewed

PR #143, opened 2022-05-16
add enum for sentry auto resolve age

maorfr:
fail fast for https://github.com/app-sre/qontract-reconcile/blob/96f3688c27fb164a9e398b32a26e478613d6db0a/reconcile/utils/sentry_client.py#L143-L190


schemas/app-sre/app-1.yml
```diff
@@ -356,6 +356,53 @@ properties:
                   type: string
               auto_resolve_age:
                 type: integer
+                enum:
```
fishi0x01, line 359: Can we have a comment here from where those values were taken?

fishi0x01 (approved):

PR #153, opened 2022-05-23
add schemas for query-validation

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5561

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/schema-compatibility.md

To avoid breaking tenants' automations that rely on app-interface, we need to make app-interface aware of the queries tenants are using in their automations. Differently worded, tenants should be able to declare queries they are using and request to be notified for breaking changes.

For that purpose, we will introduce a new schema, `query-validation-1`, which will include the common fields, such as `name` and `description`. In addition, it will include a `queries` field.

This field will be a list of references (paths) to resources (files under the `resources/` directory) which are a file containing a graphql query. Some examples can be found under `resources/queries`. This list will be the queries used in the tenants' automations, and that we need to avoid from breaking. This means that we only need to notify tenants for breaking changes that affect their queries.

In addition to the schema, we will create an integration (`query-validator`). This integration acts on `query-validation-1` files. The integration will be pretty straight forward: For each entry in the `queries` list, execute the query. If it fails - fail the integration.

This integration will only run within app-interface pr-check to gate merges of schema promotions that break the tenants' queries. In case this integration fails, we will contact the responsible team to make adjustments to their source code and to the queries in app-interface.

Once the tenants' source code was changed, together with the query files in app-interface (referenced from a `query-validation-1` file), AppSRE is clear to merge the (no-longer) breaking schema changes.

We should generally avoid breaking schema changes. At the minimum, changes should be backwards compatible to allow some time before introducing the breaking change. With that said, this design document comes as a result of a tenant request. The tenant requested 1-2 working days, so we will use that as the SLA for tenants at first.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2273,6 +2273,20 @@
   - { name: query, type: string }
   - { name: queries, type: string, isList: true }
 
+- name: QueryValidationQueryItem_v1
+  fields:
+  - { name: path, type: string, isRequired: true }
+
+- name: QueryValidation_v1
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: labels, type: json }
+  - { name: name, type: string, isRequired: true, isUnique: true, isSearchable: true }
+  - { name: description, type: string, isRequired: true }
+  - { name: escalationPolicy, type: AppEscalationPolicy_v1, isRequired: true }
+  - { name: queries, type: QueryValidationQueryItem_v1, isList: true, isRequired: true }
+
 - name: UnleashNotifications_v1
   fields:
   - { name: slack, type: SlackOutput_v1, isList: true }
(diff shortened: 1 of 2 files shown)


schemas/app-interface/query-validation-1.yml
```diff
@@ -0,0 +1,41 @@
...
+    "$ref": "/common-1.json#/definitions/labels"
+  name:
+    type: string
+    description: unique name for query validation
+  description:
+    type: string
+    description: description of the queries included in the file
+  escalationPolicy:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef": "/app-sre/escalation-policy-1.yml"
```
geoberle, line 22: ❤️

graphql-schemas/schema.yml
```diff
@@ -2273,6 +2273,20 @@
   - { name: query, type: string }
   - { name: queries, type: string, isList: true }
 
+- name: QueryValidationQueryItem_v1
```
geoberle, line 2276: nit: would it make sense to detach the QueryItem from the validation context by just naming it QontractQuery? this way queries could be referenced in a type safe manner from other places as well

maorfr: renamed in https://github.com/app-sre/qontract-schemas/pull/153/commits/c9023097a9d40060159ede5130d493538bb89a5a

geoberle (approved):
LGTM

PR #160, opened 2022-05-30
explicit resource reference metadata in qontract-schema data files

geoberle:
Introduce a common type declaration `/common-1.json#/definitions/resourceref` that is compatible with `string` but represents a path to a resource.

```yaml
---
"$schema": /metaschema-1.json
version: '1.0'
type: object
...
properties:
  ...
  path:
    "$ref": "/common-1.json#/definitions/resourceref"
```

This information is used in `qontract-validator` to verify resource references and enables fast fail for PR checks during bundle creation. At the same time it is compatible with current data.

Additionally add a property `isResourceRef = true` to the respective field in the graphql schema. This can leveraged during schema introspection.

During bundle creation, `resources.backrefs` will be filled with a list of data files that reference a resource.

References:
* design doc https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/qontract-schema-resource-references.md
* https://issues.redhat.com/browse/APPSRE-5629

Enables https://github.com/app-sre/qontract-validator/pull/37

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -368,7 +368,7 @@
   - { name: instance, type: JenkinsInstance_v1, isRequired: true }
   - { name: type, type: string, isRequired: true }
   - { name: config, type: json }
-  - { name: config_path, type: string }
+  - { name: config_path, type: string, isResourceRef: true }
 
 - name: JiraServer_v1
   fields:
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -262,6 +262,14 @@
   - { name: content, type: string, isRequired: true }
   - { name: sha256sum, type: string, isRequired: true }
   - { name: schema, type: string }
+  - { name: backrefs, type: ResourceBackref_v1, isList: true }
+
+- name: ResourceBackref_v1
+  fields:
+  - { name: path, type: string, isRequired: true }
+  - { name: datafileSchema, type: string, isRequired: true }
+  - { name: type, type: string, isRequired: true }
+  - { name: jsonpath, type: string, isRequired: true }
 
 - name: VaultSecret_v1
   fields:
@@ -998,6 +1006,7 @@
   - { name: alternate_names, type: string, isList: true }
 
 - name: NamespaceOpenshiftResource_v1
+  datafile: /openshift/openshift-resource-1.yml
   isInterface: true
   interfaceResolve:
     strategy: fieldMap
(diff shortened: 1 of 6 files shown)


bhushanthakur93:
While I was working on GraphQL schema validation, I had come across `synthetic` attribute. Based on the implementation in qontract-server it looks like it is also being used to derive backrefs, although references are not explicit and derived during query evaluation. It's also utilized by few types. I'm curious if we could achieve the same result through this or if there are any shortcomings with that attribute given the problem we are trying to solve.

maorfr:
please hold for #172

geoberle:
> While I was working on GraphQL schema validation, I had come across `synthetic` attribute. Based on the implementation in qontract-server it looks like it is also being used to derive backrefs, although references are not explicit and derived during query evaluation. It's also utilized by few types. I'm curious if we could achieve the same result through this or if there are any shortcomings with that attribute given the problem we are trying to solve.

the synthetic feature operators on a known schema where potential types and their references are explicit. for resources there is sadly no such thing. all resources are the same and can be referenced from anywhere by a simple path string. that makes backreferencing during query time very expensive. with this approach, backrefs are calculated during bundling, which also allows checking for stale references and failing the bundle validation if that is the case.

bhushanthakur93:
> > While I was working on GraphQL schema validation, I had come across `synthetic` attribute. Based on the implementation in qontract-server it looks like it is also being used to derive backrefs, although references are not explicit and derived during query evaluation. It's also utilized by few types. I'm curious if we could achieve the same result through this or if there are any shortcomings with that attribute given the problem we are trying to solve.
> 
> the synthetic feature operates on a known schema where potential types and their references are explicit. for resources there is sadly no such thing. all resources are the same and can be referenced from anywhere by a simple path string. that makes backreferencing during query time very expensive. with this approach, backrefs are calculated during bundling, which also allows checking for stale references and failing the bundle validation if that is the case.
Yeah, I agree the query time can get expensive especially if a query has multiple attribute synthetic attributes. Each time it would run through all the keys in `data` and search them. JSON bundle doesn't seem to be a right fit with this use case. A graph representation would suit better and you can pretty much do bidirectional graph for refs and backrefs. Even a directional graph can be transformed in constant time to have backrefs. 
> 
> nevertheless i would love the get rid of resources and replace them with a schema that simply wraps the content. this way we can at least leverage several validation features (validator checks on refs) or query features (synthetics).
This is something I was thinking about as well. Although I can't imagine the no of json schemas we would end up with if you also consider adding openshift resources on top. Maybe a simpler schema that captures the essence  and application logic that transforms that simple schema into a complex one (with sane defaults) could be explored. That will also limit what tenants could do and we could introduce best practices on top of that.

geoberle:
@bhushanthakur93 while i fully agree this discussion needs to be held, i would still love to move on with this MR because it will give us some safety in the meantime, e.g. stale ref checking & allowing MR checks to find out about the involved schema.

bhushanthakur93:
> @bhushanthakur93 while i fully agree this discussion needs to be held, i would still love to move on with this MR because it will give us some safety in the meantime, e.g. stale ref checking & allowing MR checks to find out about the involved schema.

@geoberle Yep, that is out of scope for this MR.

maorfr (approved):

PR #166, opened 2022-06-03
APPSRE-5552 allow the ocm-clusters integration to send slack messages

patjlm:


schemas/dependencies/slack-workspace-1.yml
```diff
@@ -56,6 +56,7 @@ properties:
           - sentry-helper
           - slack-sender
           - openshift-upgrade-watcher
+          - ocm-clusters
```
rporres, line 59: as discussed in https://github.com/app-sre/qontract-reconcile/pull/2452, it could make more sense to send these message from `ocm-upgrade-watcher`.

patjlm:
I will close this for a new implementation

PR #168, opened 2022-06-08
fix s3 schemas for external resources

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5726

depends on https://github.com/app-sre/qontract-validator/pull/38

following the validator PR, we can define `oneOf` for `refs`.

we also move the s3 definition into terraform-resource-2 because it can not be reused (terraform-resources-2 does not require an `account`).


geoberle (approved):
LGTM

PR #174, opened 2022-06-23
Add imap config to app-interface-settings - APPSRE-5894

chassing:


schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -52,6 +52,12 @@ properties:
         type: string
       credentials:
         "$ref": "/common-1.json#/definitions/vaultSecret"
+  imap:
+    type: object
+    additionalProperties: false
+    properties:
```
geoberle, line 58: since you made `credentials` required in the graphql schema, you can should add a `required` section here as well

schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -50,8 +50,25 @@ properties:
     properties:
       mailAddress:
         type: string
+      timeout:
+        type: integer
       credentials:
         "$ref": "/common-1.json#/definitions/vaultSecret"
+    required:
+    - mailAddress
+    - timeout
```
rporres, line 59: any particular reason for making this a requirement or should the code provide the current value as in case there's nothing here?

chassing: I've made these required because they are also required in the graphql schema. So far, the code doesn't have any defaults, and I think it is better to set it explicitly.

rporres: Just talking about `timeout`. I agree `mailAddress` being mandatory

chassing: I see; there are pros and cons as usual. I will make it non-required and set 30 seconds as the default

rporres (approved):

PR #175, opened 2022-06-23
Delete field for SQL Query - APPSRE-3787

cubismod:
Part of [APPSRE-3787](https://issues.redhat.com/browse/APPSRE-3787), this adds a `delete:` parameter for CronJob SQL queries so that they can be removed after being created in an automated manner.


bkez322 (approved):
LGTM

PR #151, opened 2022-05-19
Add cognito and api_gateway support

TGPSKI:
# Add Cognito and API Gateway schemas to support ROSA-FedRAMP effory

This MR adds support for AWS Cogntio and API Gateway resource creation via Terrascript -> Terraform, adding schemas for these provider types.

This MR depends on MR #38690 in Gitlab [merged], the design document documenting the use case and technical decisions that go into this change.

In addition, this MR will depend on updates to qontract-reconcile (https://github.com/app-sre/qontract-reconcile/pull/2409)

App-interface representation of new resources:

```
terraformResources:

- provider: cognito
  account: appsrefrp01ugw1
  identifier: rosa-cognito
  domain: cognito.fedramp.devshift.net
  sms_role_ext_id: foobar
  defaults: /terraform/cognito-1.yml

- provider: api-gateway
  account: appsrefrp01ugw1
  identifier: rosa-cognito
  api_proxy_uri: api-gateway.fedramp.devshift.net
  lb_subnet_id: subnet-abcdef123456
  defaults: /terraform/api-gateway-1.yml
```

PR #177, opened 2022-06-24
Add rosa-authenticator infrastructure support

TGPSKI:
This MR adds support for multiple terraform resource creation via Terrascript -> Terraform.

This MR depends on MR #38690 in Gitlab, the design document documenting the use case and technical decisions that go into this change.

In addition, this MR will depend on updates to qontract-schemas (https://github.com/app-sre/qontract-schemas/pull/151), forcing default value requirements for associated yaml files. 

App-interface representation of new resources:

```
terraformResources:

- provider: rosa-authenticator
  account: appsrefrp01ugw1
  identifier: rosa-cognito
  domain: cognito.fedramp.devshift.net
  sms_role_ext_id: foobar
  defaults: /terraform/rosa-authenticator-1.yml
```

### DIFF
--- a/schemas/aws/terraform-resource-2.yml
+++ b/schemas/aws/terraform-resource-2.yml
@@ -968,5 +968,38 @@ oneOf:
   required:
   - identifier
   - name
+- additionalProperties: false
+  properties:
+    provider:
+      type: string
+      enum:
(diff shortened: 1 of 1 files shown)


dwelch0 (approved):
LGTM

PR #185, opened 2022-07-06
enable defining github username for bots to end with [bot]

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4866

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -44,6 +44,10 @@
       "type": "string",
       "pattern": "^[a-z0-9][a-z0-9-]{3,63}[a-z0-9]$"
     },
+    "botIdentifier": {
+      "type": "string",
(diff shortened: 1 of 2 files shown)


schemas/common-1.json
```diff
@@ -44,6 +44,10 @@
       "type": "string",
       "pattern": "^[a-z0-9][a-z0-9-]{3,63}[a-z0-9]$"
     },
+    "botIdentifier": {
+      "type": "string",
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9-_]{0,30}[A-Za-z0-9](\\[bot\\])?$"
```
patjlm, line 49: if we want to respect still the 32 char max from `identifier`, should we reduce that `{0,30}` to account for `len("[bot]")`? This may impact some existing github usernames, but I think we're fine today: in commercial, the longest I found is 27 chars and is ending with `bot` anyways

PR #186, opened 2022-07-06
add resolveResource flag to graphql schema types

geoberle:
this `resolveResource` flag enables the resolution of resource ref fields during query time.

this change also deprecates the `isResourceRef` flag. `isResource` is metadata enough to determine if a field is a resource. the nuance to determine if a field is a direct resource or just a reference to it, can be done with the field type, e.g. string vs Resource_v1

see https://github.com/app-sre/qontract-server/pull/147 for the qontract-server change

ref: https://issues.redhat.com/browse/APPSRE-5950


bhushanthakur93 (approved):

bhushanthakur93:
LGTM for new changes.

PR #188, opened 2022-07-08
Add schema support for creating SNS topics and RDS event notification

OliviaHY:
#### What:
* Add schema support for creating SNS topics
* Add schema support for creating  RDS event subscription to SNS topics

#### Why:

We want to provide tenants the ability to be able subscribe to email notifications about their RDS instances's activity. 
Design doc [here](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/rds-upgrades-and-maintenance.md#database-maintenance-notifications).

#### Validation:
* Running terraform-resources integration through dry-run mode and check the output consistent with the yml defined in open source App Interface.
* Running the integration without dry-run mode against ter-int-dev AWS account and was able to see SNS topics and RDS event subscriptions created( and deleted) as intended.
yml that was used:
```
externalResources:
  - provider: aws
    provisioner:
      $ref: /aws/app-int-example/account.yml
    resources:
    - provider: rds
      identifier: app-int-example-01
      defaults: /terraform/rds/free-tier-1.yml
      output_resource_name: db-creds
      overrides:
        publicly_accessible: true
      event_notifications:
        - destination: test-sns-1
          source_type: db-instance
          event_categories: 
            - creation 
            - deletion
    - provider: sns
      identifier: test-sns-1
      defaults: /terraform/sns/sns-1.yml
      subscriptions:
        - protocol: email
          endpoint: foo@redhat.com
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -995,6 +995,12 @@ confs:
   - { name: filter_prefix, type: string}
   - { name: filter_suffix, type: string}
 
+- name: AWSRDSEventNotification_v1
+  fields:
+  - { name: destination, type: string, isRequired: true}
+  - { name: source_type, type: string}
+  - { name: event_categories, type: string, isList: true}
+
 - name: TerraformState_v1
   isInterface: true
   interfaceResolve:
@@ -1140,6 +1146,7 @@ confs:
       secrets-manager-service-account: NamespaceTerraformResourceSecretsManagerServiceAccount_v1
       aws-iam-role: NamespaceTerraformResourceRole_v1
       sqs: NamespaceTerraformResourceSQS_v1
+      sns: NamespaceTerraformResourceSNSTopic_v1
       dynamodb: NamespaceTerraformResourceDynamoDB_v1
       ecr: NamespaceTerraformResourceECR_v1
       s3-cloudfront: NamespaceTerraformResourceS3CloudFront_v1
@@ -1270,6 +1277,7 @@ confs:
   - { name: replica_source, type: string }
   - { name: ca_cert, type: VaultSecret_v1 }
   - { name: annotations, type: json }
+  - { name: event_notifications, type: AWSRDSEventNotification_v1, isList: true}
 
 - name: NamespaceTerraformResourceS3_v1
   interface: NamespaceTerraformResourceAWS_v1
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -995,6 +995,13 @@ confs:
   - { name: filter_prefix, type: string}
   - { name: filter_suffix, type: string}
 
+- name: AWSRDSEventNotification_v1
+  fields:
+  - { name: destination, type: string, isRequired: true}
+  - { name: source_type, type: string}
+  - { name: event_categories, type: string, isList: true}
+  - { name: enabled, type: boolean}
```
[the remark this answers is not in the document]
OliviaHY: No particular reason just different approach. I can remove this.

schemas/aws/terraform-resource-2.yml
```diff
@@ -454,6 +465,28 @@ oneOf:
...
+          source_type:
+            type: string
+            enum:
+            - db-instance
+            - db-security-group
+            - db-parameter-group
+            - db-snapshot
+            - db-cluster
+            - db-cluster-snapshot
+          event_categories:
```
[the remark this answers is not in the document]
OliviaHY: So one of each source_type will have its own event_categories. Are you suggesting get all the available categories and have them as enum? I was thinking this can be handled by the terraform layer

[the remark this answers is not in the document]
steveteahan: That's true... I was mostly thinking about the db-instance use case because that's our primary use case at the moment. It probably would be a bit more complicated than necessary to enumerate all of them, so this is probably fine for now.

graphql-schemas/schema.yml
```diff
@@ -1362,6 +1370,25 @@ confs:
...
+- name: NamespaceTerraformResourceSNSSubscription_v1
+  fields:
+  - { name: protocol, type: string, isRequired: true }
+  - { name: endpoint, type: string, isRequired: true}
+
+- name: NamespaceTerraformResourceSNSTopic_v1
+  interface: NamespaceTerraformResourceAWS_v1
+  fields:
+  - { name: provider, type: string, isRequired: true }
+  - { name: defaults, type: string, isResource: true, isRequired: true }
```
bhushanthakur93, line 1382: You might want to make underlying json-schema consistent with GraphQL schema in terms of requirement of the fields. Otherwise bundle validation will succeed but query evaluation will lead into error as the field is NonNull with `isRequired` attribute in GraphQL world. 

I've documented this in SCHEMA.md if you need more clarification with it. 

Otherwise LGTM.

bhushanthakur93 (approved):
Overall LGTM

graphql-schemas/schema.yml
OliviaHY: So I actually wanted to make this field not required, but was getting error. There is no way around it since the interface requires it right?

bhushanthakur93: Correct. You can make this GraphQL schema consistent by adding `defaults` to `required` list in json-schema.

schemas/aws/terraform-resource-2.yml
```diff
@@ -541,6 +572,41 @@ oneOf:
...
+        type: object
+        properties:
+          protocol:
+            type: string
+            enum:
+            - email
+          endpoint:
+            type: string
+  required:
+  - identifier
```
bhushanthakur93, line 609: You can add those required fields here.

graphql-schemas/schema.yml
bhushanthakur93: Correct.
For some reason, github added a duplicate comment and when I removed it, it got rid of my comment here. 

Anyways, resolving this now.

PR #192, opened 2022-07-15
(fedramp) rosa-authenticator: add new certificate_arn and domain_name properties

bkez322:
These are to manage domain-related terraform resources for the API-gateway.

### DIFF
--- a/schemas/aws/terraform-resource-2.yml
+++ b/schemas/aws/terraform-resource-2.yml
@@ -142,6 +142,10 @@ properties:
     type: string
   vpc_arn:
     type: string
+  vpc_arn:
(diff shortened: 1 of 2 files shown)


schemas/aws/terraform-resource-2.yml
```diff
@@ -142,6 +142,10 @@ properties:
     type: string
   vpc_arn:
     type: string
+  vpc_arn:
```
dwelch0, line 145: duplicate

dwelch0 (commented):
remove dup then LGTM

dwelch0 (approved):

PR #199, opened 2022-07-19
extract saas file target to separate schema file

maorfr:
cleaner on the eyes. saas file is long and complex enough as is.

and who knows, maybe we'll find a way to re-use targets...


geoberle (approved):
LGTM - ❤️ the mysterious comment 👍

maorfr:
https://issues.redhat.com/browse/APPSRE-6028

PR #201, opened 2022-07-20
remove unused definitions

maorfr:
were used with saas-file-v1.


maorfr:
@fishi0x01 PTAL

fishi0x01 (approved):
If those were exclusively used by `saas_file_v1`, then it should be safe to remove them :+1:

PR #213, opened 2022-08-02
resource and datafile ownership

geoberle:
introduce a generic interface `DatafileObject_v1` all datafile types automatically implement (in additional to potential other interfaces they declare). the assignment of the interface to datafile types happens automatically during bundle loading in `qontract-server`. object resolution during query time is achieved by the `interfaceResolve.strategy = schema` strategy based on the `datafile` defined for a type. this type enables mixed-type fields in graphql types where various types can be assigned.

this generic interface can be used to collect all kinds of objects in a list. 

first usecase will be declaring ownership for datafiles.

```yaml
---
$schema: /access/role-1.yml
...
self_service:
- description: AppSRE services
  datafiles:
  - $ref: /services/github-mirror/cicd/deploy.yaml
  - $ref: /services/github-mirror/cicd/test.yaml
  [other things like changetypes will go here]
  resources:
  - /app-sre/app-interface-production/qontract-api.configmap.yaml
  [other things like changetypes will go here]
```

and make them queryable like this

```
{
  roles_v1 {
    self_service {
      description
      datafiles {
        path
      }
      resources
    }
  }
}
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -283,6 +283,14 @@ confs:
   - { name: type, type: string, isRequired: true }
   - { name: jsonpath, type: string, isRequired: true }
 
+- name: DatafileObject_v1
+  isInterface: true
+  interfaceResolve:
+    strategy: schema
+  fields:
+  - { name: path, type: string, isRequired: true }
+  - { name: schema, type: string, isRequired: true }
+
 - name: VaultSecret_v1
   fields:
   - { name: path, type: string, isRequired: true }
@@ -1829,6 +1837,7 @@ confs:
       subAttr: environment
 
 - name: App_v1
+  datafile: /app-sre/app-1.yml
   fields:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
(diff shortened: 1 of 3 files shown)


graphql-schemas/schema.yml
```diff
@@ -2346,6 +2356,7 @@ confs:
   - { name: sentry_roles, type: SentryRole_v1, isList: true }
   - { name: sendgrid_accounts, type: SendGridAccount_v1, isList: true }
   - { name: owned_saas_files, type: SaasFile_v2, isList: true }
+  - { name: self_service, type: SelfServiceConfig_v1, isList: true }
```
bhushanthakur93, line 2359: Can you explain reason behind this field being a list?

geoberle: two reasons:
* for structuring sets of configs (each one has a comment)
* each SelfServiceConfig_v1 will reference one ChangeType_v1

have a look at the design doc "design doc - describing changes in app-interface data" that is open for review. there you will see how `self_service` and `change_type` will interact.

bhushanthakur93: That makes sense. Thank you!

bhushanthakur93 (approved):
Let's update the schema documentation for `datafile` field in qontract-server.

PR #216, opened 2022-08-08
Add shardSpecOverride to IntegrationSpec

janboll:
Related ticket: https://issues.redhat.com/browse/APPSRE-6100

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2443,7 +2443,6 @@ confs:
 
 - name: IntegrationSpecShardSpecOverride_v1
   fields:
-  - { name: shardId, type: int }
   - { name: awsAccount, type: AWSAccount_v1 }
   - { name: imageRef, type: string }
 
@@ -2470,6 +2469,7 @@ confs:
   - { name: restartPolicy, type: string }
   - { name: successfulJobHistoryLimit, type: int }
   - { name: failedJobHistoryLimit, type: int }
+  - { name: imageRef, type: string }
 
 - name: IntegrationManaged_v1
   fields:
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2444,7 +2444,7 @@ confs:
 - name: AWSShardSpecOverride_v1
   interface: IntegrationSpecShardSpecOverride_v1
   fields:
-    - { name: accountType, type: string, isRequired: true }
+    - { name: shardingStrategy, type: string, isRequired: true }
     - { name: awsAccount, type: AWSAccount_v1, isRequired: true }
     - { name: imageRef, type: string, isRequired: true }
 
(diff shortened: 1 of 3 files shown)


graphql-schemas/schema.yml
```diff
@@ -2386,6 +2386,11 @@ confs:
   fields:
   - { name: slack, type: boolean }
 
+- name: IntegrationSpecShardSpecOverride_v1
```
geoberle, line 2389: from a naming perspective i would love to see AWSAccount somwhere in the name of this type, since it is an overwrite specific for per-aws-account sharding

graphql-schemas/schema.yml
```diff
@@ -2386,6 +2386,11 @@ confs:
...
+  fields:
+  - { name: account, type: AWSAccount_v1 }
+  - { name: shard, type: string }
```
geoberle, line 2392: what is the purpose of those fields? from a high level perspective, both of them seem to be a specifier for the account the overwrite is for

isn't the imageRef missing?

graphql-schemas/schema.yml
geoberle: wdyt making this an interface? the sharding strategy for an integration is pluggable

janboll: Good point, I'll take another look.

graphql-schemas/schema.yml
janboll: Right, this branch is outdated. Will push an update soon

cubismod (approved):

graphql-schemas/schema.yml
geoberle: by having an awsAccount mentioned here as mandatory. there can't be any other shard overwrites except for aws shards

janboll: At this point, we have no other sharding strategies. To be frank, adding interfaces feels like overengineering to me at this point. Shouldn't it be easy to add later on?

rporres: Consistency wise, this also screams interface to me...

janboll: Hm, okay. I'll take a look into rewriting it to use interfaces

janboll:
@geoberle PTAL

graphql-schemas/schema.yml
rporres: I'll let @geoberle chime in. I understood `shardingStrategy` would be the deciding factor here, not `accountType`

graphql-schemas/schema.yml
```diff
@@ -2441,6 +2441,24 @@ confs:
...
+  fields:
+    - { name: accountType, type: string, isRequired: true }
+    - { name: awsAccount, type: AWSAccount_v1, isRequired: true }
+    - { name: imageRef, type: string, isRequired: true }
+
+- name: IntegrationSpecShardSpecOverride_v1
+  isInterface: true
+  interfaceResolve:
+    strategy: fieldMap
+    field: accountType
```
geoberle, line 2455: another sharding strategies we will have in the future is going to be `cluster` so i wonder if this discriminator field could be named `shardingStrategy` instead to cover the more generic intend.

janboll: Rename the fields but keep the interface as is, right?

geoberle: yes

janboll: okay, done

geoberle (approved):

rporres (approved):

maorfr (approved):

geoberle (approved):

PR #217, opened 2022-08-08
[integration] early exit feature for pr-check dry runs

geoberle:
the flag `/app-sre/integration-1.yml#pr_check.early_exit` controls if an integration will exit early from the potentially long running PR check dry-run if it can detect that there has been no change in desired state. the integration needs to support early exit, otherwise the flag has no effect.

q-r change: https://github.com/app-sre/qontract-reconcile/pull/2672
part of: https://issues.redhat.com/browse/APPSRE-6098

```yaml
$schema: /app-sre/integration-1.yml

name: terraform-vpc-peerings

pr_check:
  cmd: terraform-vpc-peerings
  early_exit: true
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2380,6 +2380,7 @@ confs:
   - { name: always_run, type: boolean }
   - { name: no_validate_schemas, type: boolean }
   - { name: run_for_valid_saas_file_changes, type: boolean }
+  - { name: early_exit, type: boolean }
 
 - name: IntegrationSpecExtraEnv_v1
   fields:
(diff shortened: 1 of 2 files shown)


maorfr (approved):

chassing (approved):
LGTM

patjlm:
LGTM.
But since the early exit will only be active once the integration supports the check (it has the function defined), do we really need that flag in the schema ?

geoberle:
> LGTM. But since the early exit will only be active once the integration supports the check (it has the function defined), do we really need that flag in the schema ?

sry i merged the moment your commented :(

i think it is good to have a feature toggle, e.g. we notice a bug in an integrations early-exit desired state function - the integration exits early even though there were relevant changes. in such a situation we disable early exit for that integration until the bug is fixed.

PR #218, opened 2022-08-09
Introduce Cloudflare provider & resources

jfchevrette:
Adds multiple schemas to support Cloudflare resources

Schemas added:
- CloudflareAccount_v1
- CloudflareZoneArgo_v1
- CloudflareZoneRecord_v1
- CloudflareZoneWorker_v1
- CloudflareZoneWorkerScript_v1
- Cloudflare provider for NamespaceExternalResource_v1

Part of: https://issues.redhat.com/browse/APPSRE-5715

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1154,6 +1154,7 @@ confs:
     field: provider
     fieldMap:
       zone: NamespaceTerraformResourceCloudflareZone_v1
+      worker_script: NamespaceTerraformResourceCloudflareWorkerScript_v1
   fields:
   - { name: provider, type: string, isRequired: true }
 
@@ -1161,6 +1162,7 @@ confs:
   interface: NamespaceTerraformResourceCloudflare_v1
   fields:
   - { name: provider, type: string, isRequired: true }
+  - { name: identifier, type: string, isRequired: true }
   - { name: zone, type: string, isRequired: true }
   - { name: plan, type: string }
   - { name: type, type: string }
@@ -1186,10 +1188,13 @@ confs:
   fields:
   - { name: identifier, type: string, isRequired: true }
   - { name: pattern, type: string, isRequired: true }
-  - { name: script, type: CloudflareZoneWorkerScript_v1, isRequired: true }
+  - { name: script_name, type: string, isRequired: true }
 
-- name: CloudflareZoneWorkerScript_v1
+- name: NamespaceTerraformResourceCloudflareWorkerScript_v1
+  interface: NamespaceTerraformResourceCloudflare_v1
   fields:
+  - { name: provider, type: string, isRequired: true }
+  - { name: identifier, type: string, isRequired: true }
   - { name: name, type: string, isRequired: true }
   - { name: content_from_github, type: CloudflareZoneWorkerScriptContentFromGithub_v1 }
   - { name: vars, type: CloudflareZoneWorkerScriptVars_v1, isList: true }
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1165,8 +1165,14 @@ confs:
   - { name: plan, type: string }
   - { name: type, type: string }
   - { name: settings, type: json }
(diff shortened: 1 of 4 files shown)


schemas/cloudflare/terraform-resource-1.yml
```diff
@@ -0,0 +1,147 @@
...
+          script:
+            type: object
+            additionalProperties: false
+            properties:
+              name:
+                type: string
+              content_inline:
+                type: string
+              content_file:
+                "$ref": "/common-1.json#/definitions/resourceref"
```
maorfr, line 132: i think it would be preferable that this file is stored in a separate git(hub) repo. this would be similar in behavior to how we manage lambdas (logs to elasticsearch).

schemas/cloudflare/account-1.yml
```diff
@@ -0,0 +1,44 @@
...
+          type: string
+        email:
+          type: string
+          format: email
+      required:
+      - name
+      - email
+  apiCredentials:
+    "$ref": "/common-1.json#/definitions/vaultSecret"
+  terraformStateAccount:
```
maorfr, line 33: is there a reason to use the account and not the state, like is done in accounts?

maorfr (commented):
really cool to see this come to life! :heart:

schemas/cloudflare/terraform-resource-1.yml
jfchevrette: I have refactored the schema and integration to get the content from a github repo/path/ref

schemas/cloudflare/account-1.yml
jfchevrette: The state files don't link back to an account, but accounts do link to a state file.

All other terraform integrations that we have implicitly use the same AWS account which is set as the provisioner. In the case of non-AWS resources (Cloudflare) we can't do that and we must somehow specify what AWS account we store the state in

Did I miss something that could make this better?

maorfr: i think you can crossref `/dependencies/terraform-state-1.yml` instead of `/aws/account-1.yml`?

jfchevrette: but `/dependencies/terraform-state-1.yml` does not reference an account, that is my point. so how do I know what account to use?

referencing `/aws/account-1.yml` I get both an account and its terraform state settings

maorfr: ok, understood.

if the automationToken and some of the terraform configuration was in the terraformState it would make sense.

sorry for the noise!

graphql-schemas/schema.yml
```diff
@@ -924,6 +924,19 @@ confs:
...
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: name, type: string, isRequired: true, isUnique: true, isSearchable: true }
+  - { name: description, type: string }
+  - { name: providerVersion, type: string, isRequired: true }
+  - { name: accountOwners, type: Owner_v1, isList: true, isRequired: true }
+  - { name: enableDeletion, type: boolean }
+  - { name: apiCredentials, type: VaultSecret_v1, isRequired: true }
+  - { name: terraformStateAccount, type: AWSAccount_v1, isRequired: true }
```
steveteahan, line 938: I know early on we had a lot of discussions about what this would look like. I thought that we landed on `automationToken` and doing things much like the AWS account state is handled. Can you just share what pushed you in this direction? I think it's a minor point for now as long as the keys are handled properly (no chance of overlapping keys, which looks like is the case from q-r code).

steveteahan (approved):
LGTM

graphql-schemas/schema.yml
jfchevrette: I chose to link to an AWS account (and named the key terraformStateAccount) because the integration needs both the `automationToken` and `terraformState` from the AWS account. I could have had 2 fields that links to either instead but having just 1 feels right and I cannot think of a case where we would need either from different accounts

PR #221, opened 2022-08-15
add field for ebs_options

BumbleFeng:


maorfr:
are these fields handled in qontract-reconcile?

BumbleFeng:
No, I thought they are handled. I will update q-r for it.

PR #225, opened 2022-08-18
add support to enable query support

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6098

to speed up openshift-resources (and prometheus-rules-tester) early-exit, we need to skip as many resources as possible.

resources with dynamic content (generated based on graphql query results) must be checked, but only them.
with this field, we will know exactly the ones to check. we will add a validation that if the query() function is used, this option must be set to true.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1062,6 +1062,7 @@ confs:
   - { name: validate_json, type: boolean }
   - { name: validate_alertmanager_config, type: boolean }
   - { name: alertmanager_config_key, type: string }
+  - { name: enable_query_support, type: boolean }
 
 - name: NamespaceOpenshiftResourceResourceTemplate_v1
   interface: NamespaceOpenshiftResource_v1
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1073,6 +1073,7 @@ confs:
   - { name: variables, type: json }
   - { name: validate_alertmanager_config, type: boolean }
   - { name: alertmanager_config_key, type: string }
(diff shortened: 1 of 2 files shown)


chassing (approved):
LGTM

rporres:
Can you describe with an example what you're proposing? Will we need to add `enable_query_support` everywhere?

rporres:
Ok, now I understand it better, thx for the example.

The main problem I see with this is maintainability. If we still need this after the resource reference work that is to come, I would make it in a way that it is not possible to make a mistake or forget about this, e.g. with a new resource type, something like (awful name) `resource-template-query` or whatever.

geoberle:
> Ok, now I understand it better, thx for the example.
> 
> The main problem I see with this is maintainability. If we still need this after the resource reference work that is to come, I would make it in a way that it is not possible to make a mistake or forget about this, e.g. with a new resource type, something like (awful name) `resource-template-query` or whatever.

@rporres wdyt about failing a PR check if the query function is used while this flag is not enabled?

rporres:
Yes, but why is that preferable to what I'm proposing? I understand that it would help early exit too but you would need to introduce a new check in the integration.

geoberle:
yes avoiding a new resource type (`provider`) would be the main benefit.
i don't know the story why there is `resource` and `resource-template`. they share a lot of common functionality except for the part that is handling templating. considering that i would love to avoid yet another type that is merely a subtype for templating

rporres:
Fine by me. The check of the new flag would come after the early exit so it should not make things much worse. I don't super-love the idea of adding a strange flag but I can see the benefits.

PR #236, opened 2022-09-08
introduce saas file deprecation

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6021

saas file deprecation indicates a saas file is being deprecated. with qontract-reconcile implementation, this field being true may mean that a saas file can no longer receive updates, and users will be forced to migrate resources out of it. at the very least, we can output a deprecation message.


maorfr:
example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/47029

PR #237, opened 2022-09-08
unique and identifiable objects within lists

geoberle:
finding differences in datafiles is a complex task, especially when dealing with lists of objects, where objects don't have clear identities. diff libraries like pythons deepdiff do a great job at this, using heuristics to find out if an objects in a list was just moved to another place or if the object has really changed significantly. diff libraries can be further supported in yielding better results, by providing hints about what properties of an object have key characteristics within the context of the current list.

our current gql type declaration syntax supports uniqueness on single properties globally within a bundle with the `isUnique` flag.

within certain contexts, that is too restrictive or a single property can not even be used to establish uniqueness. therefore this PR proposes a new flag `isContextUnique` that can be put on multiple properties of a type. doing so has no influence on global uniqueness within the bundle, so the functionality of `isUnique` is not touched by that. `isContextUnique` will do two things:

- qontract-validator will fail the bundling step if an object is not unique within a list
- qontract-validator will add a hidden identity field to the datafiles of a bundle which to represent the uniquness boild down to a hash value

especially the hidden field can be used to improve diff extraction from bundle changes in the context of the granular permission model and the `change-owner` integration driving that process.

(!) some sort of uniqueness is required on saas file resource template targets to ensure good results on exposing differences. the namespace alone is not sufficient in all cases, so this PR also adds an optional name field to is used for uniqueness as well

References:

- change-owner granular permission model - https://github.com/app-sre/qontract-reconcile/pull/2772
- deepdiff - https://zepworks.com/deepdiff/5.8.1/index.html
- qontract-validator PR - https://github.com/app-sre/qontract-validator/pull/43
- schema docs update PR - https://github.com/app-sre/qontract-server/pull/163

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1063,8 +1063,8 @@ confs:
 - name: NamespaceOpenshiftResourceResource_v1
   interface: NamespaceOpenshiftResource_v1
   fields:
-  - { name: provider, type: string, isRequired: true }
-  - { name: path, type: string, isRequired: true, isResource: true, resolveResource: true }
+  - { name: provider, type: string, isRequired: true, isContextUnique: true }
+  - { name: path, type: string, isRequired: true, isResource: true, resolveResource: true, isContextUnique: true }
   - { name: validate_json, type: boolean }
   - { name: validate_alertmanager_config, type: boolean }
   - { name: alertmanager_config_key, type: string }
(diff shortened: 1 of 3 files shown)


maorfr (approved):
LGTM

bhushanthakur93:
LGTM! 

It will be nice to see some kind of native support for object hash in generated types. A similar example would be how ORM like hibernate do this through `equals` and `hashCode` methods in Java.

bhushanthakur93 (approved):

PR #238, opened 2022-09-09
aws: Allow 'release' as a ref for the asg image

Gundersanne:
This upstream ref is resolved and used to find the image with a tag containing the resolved ref.

---

Context:
Our upstream now contains the https://github.com/osbuild/osbuild-composer/tree/release ref, which tracks the commits on main for which CI has passed. Due to the size of our CI having a proper merge queue is impractical, so unrebased PRs break composer's main branch now and then. But with this band-aid in place, stage can track upstream again (mostly anyway, currently the ref is pinned).

### DIFF
--- a/schemas/aws/terraform-resource-2.yml
+++ b/schemas/aws/terraform-resource-2.yml
@@ -1026,7 +1026,7 @@ oneOf:
             ref:
               description: commit or branch name
               type: string
-              pattern: '^([0-9a-f]{40}|master|main)$'
+              pattern: '^([0-9a-f]{40}|master|main|release)$'
             upstream:
               description: jenkins instance and job to wait for
               type: object


maorfr:
https://github.com/app-sre/qontract-schemas/pull/120#issuecomment-1110563130

Gundersanne:
> [#120 (comment)](https://github.com/app-sre/qontract-schemas/pull/120#issuecomment-1110563130)

Right, we wouldn't use the release branch in production; just stage, production would always be pinned. Would a different name work?

maorfr:
Have you considered main/master? 😅

maorfr:
I'd add a `ci` branch "under" main.

Gundersanne:
> Have you considered main/master? sweat_smile

Ideally yes, but one [run of the integration suite](https://gitlab.com/redhat/services/products/image-builder/ci/osbuild-composer/-/pipelines/633373340) takes up to 3 hours, and there's limited room for improvement there currently.

So then I'd prefer to keep pinning  specific refs due to merge queues being too impractical for [composer](https://github.com/osbuild/osbuild-composer) currently. 

> I'd add a ci branch "under" main.

Might work, will mark as draft in the meantime.

Gundersanne:
How about `main-passed-ci` ; just very descriptive?

maorfr:
we've been keeping this very clean, and would do a lot to keep it that way :sweat_smile:

PR #243, opened 2022-09-16
APPSRE-6153 reduce optionals

chassing:


geoberle (approved):
since our Apollo GQL resolver return an empty list when nothing is found for data datafile schema, this change works nicely and i would even suggest we do this for all queries!

chassing:
Should I add `isRequired: true` for all (`synthetic`s and all in `Query`)?

geoberle:
> Should I add `isRequired: true` for all (`synthetic`s and all in `Query`)?

lets get team consensus first. but that should not stop you from merging this

PR #244, opened 2022-09-18
extract self-service config to its own schema

maorfr:
this will allow us to preserve the ability to query the referencing configs from the context of an owned file, like we have today from saas files to roles.

saas file -> self-service config -> role.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2426,6 +2426,7 @@ confs:
 
 - name: SelfServiceConfig_v1
   fields:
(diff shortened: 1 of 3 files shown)


maorfr:
example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/47826

maorfr:
going to reshape this following https://github.com/app-sre/qontract-server/pull/165

geoberle (commented):
looks good. i will have another look when you are done "reshaping"

PR #245, opened 2022-09-19
enable disabling of terraform-vpc-peerings at an account level

maorfr:


maorfr:
required for https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/46818

PR #252, opened 2022-09-22
add enum for upgrade policy allowed workloads

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4709

with this list in place, we can add a validation that cluster upgrade policies are only using a workload name from this list.
this will help us avoid adding a "hard coded" enum to the schemas repo, which is AppSRE specific.

### DIFF
--- a/schemas/openshift/openshift-cluster-manager-1.yml
+++ b/schemas/openshift/openshift-cluster-manager-1.yml
@@ -30,6 +30,11 @@ properties:
     type: array
     items:
       type: string
+  upgradePolicyAllowedWorkloads:
+    description: List of workloads that are allowed to be defined as cluster upgrade policy mutexes
+    type: array
+    items:
+      type: string
 required:
 - "$schema"
 - labels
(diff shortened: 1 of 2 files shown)


maorfr:
example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/48193

patjlm (approved):
Makes sense.
This enum would then be enforced by the integration ?

schemas/openshift/openshift-cluster-manager-1.yml
```diff
@@ -30,6 +30,11 @@ properties:
     type: array
     items:
       type: string
+  upgradePolicyAllowedWorkloads:
+    description: List of workloads that are allowed to be defined as cluster upgrade policy mutexes
```
patjlm, line 34:
```suggestion
    description: List of workloads that are allowed to be defined as cluster upgrade policy workload
```

maorfr:
i'm still thinking of how to enforce it. stay tuned!

maorfr:
following up in #255

PR #253, opened 2022-09-22
APPSRE-6321 Add ShardSpecOverride to MR checks

janboll:
Make it possible to override the image for Merge Request checks

### DIFF
--- a/schemas/app-sre/integration-1.yml
+++ b/schemas/app-sre/integration-1.yml
@@ -55,6 +55,24 @@ properties:
         type: boolean
       early_exit:
         type: boolean
+      shardSpecOverride:
+        type: array
+        items:
+          type: object
+          additionalProperties: false
+          properties:
+            shardingStrategy:
+              type: string
+              enum:
+                - per-aws-account
+            awsAccount:
+              "$ref": "/common-1.json#/definitions/crossref"
+              "$schemaRef": "/aws/account-1.yml"
+            imageRef:
+              type: string
+          required:
+          - awsAccount
+          - imageRef
     required:
     - cmd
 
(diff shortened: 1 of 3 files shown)


schemas/app-sre/integration-1.yml
```diff
@@ -55,6 +55,24 @@ properties:
...
+              enum:
+                - per-aws-account
+            awsAccount:
+              "$ref": "/common-1.json#/definitions/crossref"
+              "$schemaRef": "/aws/account-1.yml"
+            imageRef:
+              type: string
+          required:
+          - awsAccount
+          - imageRef
```
maorfr, line 75: if this is now a duplicate, i would suggest extracting it to a separate schema

janboll: Good point! Will do

janboll: I refactored the schema. However this now crashes the qontract-server. Validation works fine though.

janboll: Nevermind, was related to local image mess

cubismod (approved):

schemas/app-sre/integration-shardspec-override-1.yml
```diff
@@ -0,0 +1,26 @@
...
+      - per-aws-account
+  awsAccount:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef": "/aws/account-1.yml"
+  imageRef:
+    type: string
+
+required:
+- awsAccount
+- imageRef
```
geoberle, line 26: no end of line 😱

geoberle (approved):
nit - otherwise lgtm

schemas/app-sre/integration-shardspec-override-1.yml
janboll: added

PR #255, opened 2022-09-22
add option to validate resources with queries

maorfr:
related to https://issues.redhat.com/browse/APPSRE-4709

query-validator currently only validates that queries are executed successfully.

with the addition of `resources`, we can do everything we already can in `openshiftResources` (jinja2 for example). with this addition, we can validate the actual data, and not only the query results.

logic added in https://github.com/app-sre/qontract-reconcile/pull/2807


maorfr:
example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/48204

patjlm (approved):
LGTM, nice!

PR #259, opened 2022-09-29
add option to trigger deployment following an image push

maorfr:
part of https://issues.redhat.com/browse/APPSRE-5799

with this schema, we can point a target at an image, like we do for a jenkins job. whenever there is a change pushed to the ref, and an image is pushed - trigger!


maorfr:
example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/48679

schemas/app-sre/saas-file-target-1.yml
```diff
@@ -73,6 +73,18 @@ properties:
     required:
     - instance
     - name
+  image:
```
rporres, line 76: Can you add some nice descriptions here documenting the new fields for future reference?

PR #261, opened 2022-10-02
add saas file target as an allowed schema reference of datafiles

maorfr:
enable usage of change types to define ownership over saas file target files.

ref: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/48797


maorfr:
even with this PR, we still can't define ownership of targets. when trying to add the `datafile` field to the graphql schema, we get this error:
```
Error: Interface field DatafileObject_v1.path expects type String! but SaasResourceTemplateTarget_v2.path is type String.
```

PR #263, opened 2022-10-02
add datafile field to graphql saas file target

maorfr:
this PR is supposed to fix #261 and allow usage of saas file targets as `datafiles`.
alas, a saas file target can either be inline or referenced, hence the `path` field is not mandatory.
this is causing the following error:
```
Error: Interface field DatafileObject_v1.path expects type String! but SaasResourceTemplateTarget_v2.path is type String.
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2218,13 +2218,50 @@ confs:
   - { name: hash_length, type: int }
   - { name: parameters, type: json }
   - { name: secretParameters, type: SaasSecretParameters_v1, isList: true }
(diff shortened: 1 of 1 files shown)


maorfr:
while this is passing, it's not possible to start a server.

geoberle (approved):

cubismod:
Checking to see whether my changes to PR check will reveal the error here

cubismod:
Woo! That's the type of fail I like to see!

maorfr:
there is an attempt to split a saas file to targets in https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/52175

@geoberle how do you think we should proceed?

PR #273, opened 2022-10-11
Change prometheus rule test schema for floats

saswatamcode:
This PR changes the Prometheus rule test schema `exp_samples` under `promql_expr_test` to accommodate floating points instead of just integers, as a series sample can be a float.


matej-g (approved):
:rocket:

jessicalins (approved):
:sparkler:

PR #274, opened 2022-10-14
glitchtip schemas

chassing:
This implements the schema for the qontract-reconcile [glitchtip](https://gitlab.com/glitchtip/glitchtip-backend) integration.

Ticket: [APPSRE-6260](https://issues.redhat.com/browse/APPSRE-6260)
Design Doc: [Glitchtip](https://gitlab.cee.redhat.com/cassing/app-interface/-/blob/master/docs/app-sre/design-docs/glitchtip.md)

### DIFF
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -160,6 +160,20 @@ properties:
     required:
     - readTimeout
     - connectTimeout
+  glitchtip:
+    type: object
+    additionalProperties: false
+    properties:
+      readTimeout:
+        type: integer
+      maxRetries:
+        type: integer
+      mailAddress:
+        type: string
+    required:
+    - readTimeout
+    - maxRetries
+    - mailAddress
 required:
 - "$schema"
 - labels
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -168,12 +168,8 @@ properties:
         type: integer
       maxRetries:
         type: integer
-      mailAddress:
+      mailDomain:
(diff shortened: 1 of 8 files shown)


rporres (commented):
I will come back when https://github.com/app-sre/qontract-reconcile/pull/2825 is marked as ready, but it looks great. Very nice use of synthetic attributes, even nested <3

schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -160,6 +160,20 @@ properties:
...
+  glitchtip:
+    type: object
+    additionalProperties: false
+    properties:
+      readTimeout:
+        type: integer
+      maxRetries:
+        type: integer
+      mailAddress:
+        type: string
```
rporres, line 172: Reading the code, I would call this `mailDomain`

schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -160,6 +160,20 @@ properties:
...
+    required:
+    - readTimeout
+    - maxRetries
+    - mailAddress
```
rporres, line 176: why making this mandatory instead of providing sane defaults in the code? We would avoid breaking changes...

schemas/app-interface/app-interface-settings-1.yml
chassing: you're right 🤦

schemas/app-interface/app-interface-settings-1.yml
chassing: It's not required, there are defaults in the code.

rporres (approved):

PR #278, opened 2022-10-20
change-type inherritence

maorfr:
this is a proposal to implement change-type inheritance, where one change-type can reference other change-types and get their permissions as well.

this may help us prevent duplication between change-types where we want to include the same permissions and will provide reusability of change-types.

part of: https://issues.redhat.com/browse/APPSRE-6473

### DIFF
--- a/schemas/app-interface/change-type-1.yml
+++ b/schemas/app-interface/change-type-1.yml
@@ -22,9 +22,9 @@ properties:
   disabled:
     type: boolean
     description: a disabled change-type does not have any effect but will still log into the PR check logs
-  inherrit:
+  inherit:
     type: array
-    description: inherrit the permissions of additional change types
+    description: inherit the permissions of additional change types
     items:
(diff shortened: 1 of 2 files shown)


schemas/app-interface/change-type-1.yml
```diff
@@ -22,6 +22,12 @@ properties:
   disabled:
     type: boolean
     description: a disabled change-type does not have any effect but will still log into the PR check logs
+  inherit:
+    type: array
+    description: inherit the permissions of additional change types
```
patjlm, line 27:
```suggestion
    description: inherit the changes of additional change types
```

geoberle (approved):

schemas/app-interface/change-type-1.yml
geoberle, line 27:
```suggestion
    description: inherit the changes of additional change types
```

geoberle: !

PR #289, opened 2022-10-25
add deletionApprovals for Cloudflare accounts

jfchevrette:
part of https://issues.redhat.com/browse/APPSRE-6478

This renames the `AWSAccountDeletionApproval_v1` schema to a more generic name `DeletionApproval_v1` and adds it to the `CloudflareAccount_v1` schema

The object definition for a `deletionApproval` is also extracted out of AWS and Cloudflare accounts schemas into a common definition

This is required by https://github.com/app-sre/qontract-reconcile/pull/2896

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -913,7 +913,7 @@ confs:
   - { name: automationToken, type: VaultSecret_v1, isRequired: true }
   - { name: garbageCollection, type: boolean }
   - { name: enableDeletion, type: boolean }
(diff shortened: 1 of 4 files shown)


maorfr:
you don't have to put deletionApproval in the common schema. you can put it in its own schema file and reference it.

example: https://github.com/app-sre/qontract-schemas/blob/eb2e10ba64da98f0c8b8de9bc6e17cb518233bc8/schemas/openshift/external-resource-1.yml

maorfr (approved):

PR #302, opened 2022-11-03
[ocm-upgrade-scheduler-org] enable support for dynamic cluster discovery

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6517

in order to keep `upgradePolicyClusters` up-to-date with clusters being created/deleted, we need to define conditions of clusters (ClusterDeployment labels a.k.a external configuration labels) based on which we select an upgrade policy for a discovered cluster.

the intention is to create a separate integration that will submit MRs to app-interface to update `upgradePolicyClusters`, so we can keep the logic of `ocm-upgrade-scheduler-org` unchanged. the reason is that the current "static" nature of defining clusters is a no-go for management of OSD-FM fleets.

the OSD FM requirements are more complex compared to a regular customer, which usually has a single OCM org in a single OCM environment with a `statically managed set of clusters.

### DIFF
--- a/schemas/openshift/openshift-cluster-manager-1.yml
+++ b/schemas/openshift/openshift-cluster-manager-1.yml
@@ -40,6 +40,27 @@ properties:
     type: array
     items:
       type: string
+  upgradePolicyDefaults:
+    description: List of default upgrade policies to apply to clusters by external configuration labels
+    type: array
+    items:
+      type: object
+      additionalProperties: false
+      properties:
+        name:
+          description: upgrade policy name
+          type: string
+        matchLabels:
+          description: external configuration labels to select default upgrade policy by
+          type: object
+          additionalProperties:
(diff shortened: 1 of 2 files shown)


schemas/openshift/openshift-cluster-manager-1.yml
```diff
@@ -55,6 +76,9 @@ properties:
       required:
       - name
       - upgradePolicy
+dependencies:
+  upgradePolicyDefaults:
+  - upgradePolicyClusters
```
patjlm, line 81: I don't understand this section. What does that do ?

maorfr: it means that `upgradePolicyClusters` must exist if `upgradePolicyDefaults` is defined

patjlm (approved):

PR #303, opened 2022-11-03
ROSA Allow multiple OCM environments per AWS acccount

lechuk47:


schemas/dependencies/rosa-ocm-1.yml
```diff
@@ -0,0 +1,30 @@
+---
```
rporres, line 1: Can we maintain the nice descriptions that we had in the previous file if they're still relevant?

lechuk47: The main purpose of these comments was to notice the default assignment if they were empty and that is no longer happening. These roles are a very basic ROSA concept that is described extensively in the ROSA docs and the names are self-descriptive by its names.

PR #308, opened 2022-11-08
Add attributes for gitlab sync support

dwelch0:
Adds support for specifying partitioned gitlab instance to copy repositories to. https://github.com/app-sre/git-partition-sync-producer performs the graphql query to obtain the sync configuration.

ticket: https://issues.redhat.com/browse/APPSRE-6544

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2002,11 +2002,23 @@ confs:
   fields:
   - { name: role, type: Role_v1, isRequired: true }
 
+- name: CodeComponentGitlabSync_v1
+  fields:
+  - { name: sourceProject, type: CodeComponentGitlabSyncProject_v1, isRequired: true }
+  - { name: destinationProject, type: CodeComponentGitlabSyncProject_v1, isRequired: true }
+
+- name: CodeComponentGitlabSyncProject_v1
+  fields:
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -2004,11 +2004,23 @@ confs:
...
 
+- name: CodeComponentGitlabSync_v1
+  fields:
+  - { name: sourceProject, type: CodeComponentGitlabSyncProject_v1, isRequired: true }
+  - { name: destinationProject, type: CodeComponentGitlabSyncProject_v1, isRequired: true }
+
+- name: CodeComponentGitlabSyncProject_v1
+  fields:
+  - { name: name, type: string, isRequired: true }
+  - { name: group, type: string, isRequired: true }
```
TGPSKI, line 2015: What is group here? Is this akin to "app-sre" or "service" in GitLab?

TGPSKI (approved):

graphql-schemas/schema.yml
dwelch0: correct. Within app logic, group/name are utilized to supply a gitlab PID for api requests. The alternative was specifying project ID directly but this seemed easier / more clear when reviewing

PR #312, opened 2022-11-14
Allow saas file timeouts > 60 minutes

rporres:
APPSRE-4670

Signed-off-by: Rafael Porres Molina <rporres@gmail.com>

### DIFF
--- a/schemas/app-sre/saas-file-2.yml
+++ b/schemas/app-sre/saas-file-2.yml
@@ -117,12 +117,13 @@ properties:
   compare:
     type: boolean
   timeout:
-    type: integer
-    minimum: 60
+    type: string
+    pattern: '(\d+[dhms])+'
     description: |
-      Maximum time (in minutes) the deployment job is allowed to last. In Tekton it defaults to
-      60 minutes.  We don't allow smaller times as Tekton implementation currently deletes
-      PipelineRun pods, hence destroying logs.
+      Maximum time the deployment job is allowed to last. In Tekton it defaults to
+      60 minutes. We don't allow smaller times as Tekton implementation currently deletes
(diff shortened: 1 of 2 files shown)


kwilczynski (approved):
Looks good!

schemas/app-sre/saas-file-2.yml
```diff
@@ -116,6 +116,13 @@ properties:
     description: prevent updates to a saas file
   compare:
     type: boolean
+  timeout:
```
geoberle, line 119: considering that tenants will most likely not read the schema description, i wonder if we should somehow add the unit to the property name like `timeoutInMinutes`. i know it does not look nice and i'm aware that we have various timeout fields already (most of them without doc string so who knows if it is seconds or minutes or something else) - so i leave this up to you. feel free to ignore.

i would love to have timeouts like these as a string like `timeout: 60m` but we would loose the possibility to check for `minimum` :(

geoberle: after having a look at https://github.com/app-sre/qontract-reconcile/pull/2955 i realised that you do additional checks in code as well. that would open up the possibility for the string based version that includes the unit like `pattern: "^\\d+[smhd]$"` and you would still be able to fail PR checks if the value is not considered within the valid range.

rporres: It makes sense. It'll make the feature nicer.

rporres: Fixed in https://github.com/app-sre/qontract-schemas/pull/312/commits/17fc3eac2abbe1830a3ca47c0b56f066d1102b45

geoberle (approved):
LGTM

PR #313, opened 2022-11-14
Draft: Creating new inheritVersionHistory field for SRE capabilities

suzana-nesic:
Signed-off-by: Suzana Nesic <snesic@redhat.com>


patjlm (approved):
LGTM. for the PR description, I think `production` would inherit from `stage`, not the opposite

suzana-nesic:
Yes, thank you! Was working with what was currently in a-i and didn't update to what we should expect. Will fix the description

PR #317, opened 2022-11-16
cluster.auth attribute becomes a list of objects instead of a single object

chassing:
Breaking Change: The `cluster.auth` attribute is now a list and **required**

We want to support [multiple auth backends](https://issues.redhat.com/browse/APPSRE-6374) at the same time.

* [APPSRE-6374](https://issues.redhat.com/browse/APPSRE-6374)
* [APPSRE-6554](https://issues.redhat.com/browse/APPSRE-6554)

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -793,7 +793,7 @@ confs:
   - { name: labels, type: json }
   - { name: name, type: string, isUnique: true, isSearchable: true, isRequired: true }
   - { name: description, type: string }
-  - { name: auth, type: ClusterAuth_v1, isInterface: true }
+  - { name: auth, type: ClusterAuth_v1, isInterface: true, isList: true, isRequired: true }
   - { name: observabilityNamespace, type: Namespace_v1 }
   - { name: grafanaUrl, type: string }
   - { name: consoleUrl, type: string, isRequired: true }
--- a/schemas/openshift/cluster-1.yml
+++ b/schemas/openshift/cluster-1.yml
@@ -32,49 +32,51 @@ properties:
     type: string
     format: uri
   auth:
-    type: object
-    properties:
-      service:
-        type: string
-        enum:
-        - github-org
-        - github-org-team
-        - oidc
-      org:
-        type: string
-      team:
-        type: string
-    oneOf:
-    - properties:
+    type: array
+    items:
+      type: object
+      properties:
         service:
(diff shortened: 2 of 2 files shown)


janboll (approved):

graphql-schemas/schema.yml
```diff
@@ -793,7 +793,7 @@ confs:
   - { name: labels, type: json }
   - { name: name, type: string, isUnique: true, isSearchable: true, isRequired: true }
   - { name: description, type: string }
-  - { name: auth, type: ClusterAuth_v1, isInterface: true }
+  - { name: auth, type: ClusterAuth_v1, isInterface: true, isList: true, isRequired: true }
```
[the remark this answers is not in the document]
chassing: It'll be defined (redefined) for each cluster. see [this MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/51932)

patjlm, line 796: I think we should not have `isRequired: true`. Some clusters do not define this, when we import them in app-interface and we do not manage them ourselves. For example `c-rh-c-eph`

schemas/openshift/cluster-1.yml
```diff
@@ -32,49 +32,51 @@ properties:
...
+        - service
+        - org
+        - team
+      - properties:
+          service:
+            type: string
+            enum:
+            - oidc
+        required:
+        - service
```
patjlm, line 79: does `oidc` mean auth.redhat.com? Should we name it against the provider rather than the protocol, eg `oidc-auth-redhat`?

graphql-schemas/schema.yml
chassing: I would like to have it as required because the qontract-reconcile code is much easier to read. No `auth is not None` or `for a in auth or []` everywhere. I'll add `auth: []` to those clusters without an auth ([see this MR](https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/51932))

patjlm: hmm, I see. Fair enough

schemas/openshift/cluster-1.yml
chassing: This is as it was before. I'll adapt the schemas for auth.redhat.com in a later PR

rporres (approved):

PR #318, opened 2022-11-16
add 2fa enforcement support to cloudflare account schema

jfchevrette:
[APPSRE-6571](https://issues.redhat.com/browse/APPSRE-6571)

This is the schema needed for https://github.com/app-sre/qontract-reconcile/pull/2964 which allows setting 2FA enforcement on Cloudflare accounts via the account definition file

### DIFF
--- a/schemas/cloudflare/account-1.yml
+++ b/schemas/cloudflare/account-1.yml
@@ -30,6 +30,8 @@ properties:
       - email
   apiCredentials:
     "$ref": "/common-1.json#/definitions/vaultSecret"
+  enforceTwofactor:
+    type: boolean
   terraformStateAccount:
     description: AWS Account to use for state in S3
     "$ref": "/common-1.json#/definitions/crossref"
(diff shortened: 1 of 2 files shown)


steveteahan (approved):

schemas/cloudflare/account-1.yml
```diff
@@ -30,6 +30,8 @@ properties:
       - email
   apiCredentials:
     "$ref": "/common-1.json#/definitions/vaultSecret"
+  enforceTwofactor:
```
janboll, line 33: NIT: isn't it `enforceTwoFactor` ?

jfchevrette: I converted the [terraform provider parameter](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/account) from snake_case to camelCase to match the style of the other parameters

Often times we keep the snake_case version of terraform parameters when we use them in terraform resources schemas. Here it's not really a resource definition for more of an account property that gets turned into a terraform parameter under the hood.

I don't really have a good argument either way. I think it's fine to merge as-is?

PR #319, opened 2022-11-17
override and default schema represented by same schema objects

geoberle:
all CNA assets must have a `defaults` and an `overrides` section, as a lot of the terraform-resources do. the difference is, that each CNA type declares a special `XXXConfig_v1` type that is used for both fields.

```yaml
  name: CNARDSInstance_v1
  interface: CNAsset_v1
  fields:
  - { name: provider, type: string, isRequired: true }
  - { name: identifier, type: string, isRequired: true, isUnique: true }
  - { name: name, type: string }
  - { name: defaults, type: CNARDSInstanceConfig_v1 } <--
  - { name: overrides, type: CNARDSInstanceConfig_v1 } <--
```

having defaults following a strict schema and overrides being able to override all of the defaults, makes writing testable and verifiable code a lot easier.

### DIFF
--- a/schemas/cna/null-asset-config-1.yml
+++ b/schemas/cna/null-asset-config-1.yml
@@ -1,0 +1,12 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /cna/null-asset-config-1.yml
+  addr_block:
+    type: string
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1007,10 +1007,10 @@ confs:
   - { name: provider, type: string, isRequired: true }
   - { name: identifier, type: string, isRequired: true, isUnique: true }
   - { name: aws_account, type: AWSAccount_v1, isRequired: true }
-  - { name: overrides, type: CNAAssumeRoleAssetOverrides_v1 }
-  - { name: defaults, type: string, isRequired: true, isResource: true, resolveResource: true }
+  - { name: defaults, type: CNAAssumeRoleAssetConfig_v1 }
+  - { name: overrides, type: CNAAssumeRoleAssetConfig_v1 }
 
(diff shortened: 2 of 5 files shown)


schemas/cna/null-asset-config-1.yml
```diff
@@ -0,0 +1,12 @@
...
+version: '1.0'
+type: object
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /cna/null-asset-config-1.yml
+  addr_block:
+    type: string
```
fishi0x01, line 12: Should this also have

```
required:
- "$schema"
```
?

geoberle: a type that will to be used inline and crossrefed can't require the schema field otherwise we would need to define an override like this

```yaml
  - provider: null-asset
    identifier: my-null
    overrides:
      schema: /cna/null-asset-config-1.yml
      addr_block: bla
```

geoberle: this can be observed also here https://github.com/app-sre/qontract-schemas/blob/3803f49c99dc0319226d5455c0d7ddd841881ce2/schemas/app-sre/saas-file-target-1.yml

fishi0x01: But isnt that also true for the other 2 configs [aws-assume-role-config-1.yml](https://github.com/app-sre/qontract-schemas/pull/319/files#diff-f7a9830e1528f42d08fa0d8460a8b0f874d21cdb6481f9a37b38a8d516f04f87) and [aws-rds-config-1.yml](https://github.com/app-sre/qontract-schemas/pull/319/files#diff-e1c191c537f117368689cc81263cd9f575a08eb36217ac8f42208741db3c2e0e) ? I do not understand why they are handled differently here?

geoberle: that was unintentional and basically a bug that would have surfaced when we would have started using overrides for those two asset types.

fixed in ff793ba

good catch!

fishi0x01 (approved):

PR #321, opened 2022-11-17
add 2fa enforcement support to cloudflare account schema

jfchevrette:
This is a rework of #320 which was reverted

[APPSRE-6571](https://issues.redhat.com/browse/APPSRE-6571)

This is the schema needed for https://github.com/app-sre/qontract-reconcile/pull/2964 which allows setting 2FA enforcement on Cloudflare accounts via the account definition file. A field `type` is also needed by this resource

This is required for
* https://github.com/app-sre/qontract-reconcile/pull/2970


jfchevrette:
Will merge this w/o approvals as this is the same code as in https://github.com/app-sre/qontract-schemas/pull/320 with an added field

PR #324, opened 2022-11-21
[cna] RDS overrides as json with all optional

geoberle:
revert experiment where overrides and defaults are exactly the same schema. the result was that all fields in overrides and defaults needed to be optional so they can be used in both places without making overrides mandatory.

in this PR, overrides are now all optional, follow a schema in jsonschema but none in GQL where they are just JSON

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -171,6 +171,49 @@
         "partial_outage",
         "major_outage"
       ]
+    },
+    "rdsDbInstanceClass": {
+      "type": "string",
+      "enum": [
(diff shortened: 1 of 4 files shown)


schemas/common-1.json
```diff
@@ -171,6 +171,49 @@
         "partial_outage",
         "major_outage"
       ]
+    },
+    "rdsDbInstanceClass": {
```
fishi0x01, line 175: neat :+1:

fishi0x01 (approved):

PR #325, opened 2022-11-21
upgradePolicyTemplate in OCM upgradePolicyDefaults

patjlm:
Allow to reference an upgradepolicy template from OCM `upgradePolicyDefaults`.

This avoids repeating many time the same default for different labels such as `sector` (or region, or cluster-type, ...).

https://issues.redhat.com/browse/APPSRE-6595

Example usage: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/52219/diffs


geoberle (approved):
LGTM

PR #332, opened 2022-11-28
glitchtip associate to namespace instead of app

maorfr:
the long term goal is to fetch the DSN of projects back to namespaces (and vault) to be consumed by services directly. that will prevent users from having to manually copy the DSN.


chassing (approved):

graphql-schemas/schema.yml
```diff
@@ -3006,6 +3008,11 @@ confs:
   - { name: teams, type: GlitchtipTeam_v1, isRequired: true, isList: true }
   - { name: organization, type: GlitchtipOrganization_v1, isRequired: true, isUnique: true }
 
+- name: GlitchtipProjectsReference_v1
```
geoberle, line 3011: first time it's a hack, second time it's a trick ❤️

geoberle (approved):
nice! having glitchtip projects linked to namespaces makes a lot of sense!
LGTM

PR #334, opened 2022-11-28
glitchtip queries adjustments

maorfr:
follow up on #332

- add ability to query all glitchtip projects
- make glitchtip projects reference only (no inline)
- add backrefs to find namespaces and shared resources from glitchtip project

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3009,7 +3009,10 @@ confs:
   - { name: organization, type: GlitchtipOrganization_v1, isRequired: true, isUnique: true }
 
 - name: GlitchtipProjects_v1
+  datafile: /dependencies/glitchtip-project-1.yml
   fields:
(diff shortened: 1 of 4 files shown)


graphql-schemas/schema.yml
```diff
@@ -1886,6 +1886,13 @@ confs:
   - { name: openshiftResources, type: NamespaceOpenshiftResource_v1, isList: true, isInterface: true, isRequired: true }
   - { name: openshiftServiceAccountTokens, type: ServiceAccountTokenSpec_v1, isList: true }
   - { name: glitchtipProjects, type: GlitchtipProjects_v1, isList: true }
+  - name: namespaces
```
chassing, line 1889: Why do you need a backref to namespaces?

maorfr: to understand later where to put Secrets with the DSN

chassing (approved):
LGTM

PR #341, opened 2022-11-30
OIDC IDP cluster auth object

chassing:
Enhance `cluster.auth.oidc` schema.

APPSRE-6622

### DIFF
--- a/schemas/openshift/cluster-1.yml
+++ b/schemas/openshift/cluster-1.yml
@@ -75,8 +75,32 @@ properties:
             type: string
             enum:
             - oidc
+          name:
+            type: string
+          issuer:
+            type: string
+            format: uri
+          claims:
+            email:
+              type: array
+              items:
+                type: string
+            username:
+              type: array
+              items:
+                type: string
+            name:
+              type: array
+              items:
+                type: string
+            groups:
+              type: array
+              items:
+                type: string
         required:
         - service
+        - name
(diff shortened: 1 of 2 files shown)


schemas/openshift/cluster-1.yml
```diff
@@ -75,8 +75,32 @@ properties:
...
+            type: string
+          issuer:
+            type: string
+            format: uri
+          claims:
+            email:
+              type: array
+              items:
+                type: string
+            username:
```
geoberle, line 88: do we need a preferred username config option as well?

chassing: `username` is the attribute for OCM `preferred_username`

geoberle: the ocm api provides an option to specify both - see `OpenIdClaims.preferred_username` on https://api.openshift.com/#/

thats why i thought it might be relevant. if it is not relevant in our context, feel free to ignore

chassing: `username` will be translated to `preferred_username` by the [integration](https://github.com/app-sre/qontract-reconcile/pull/3029/files#diff-035de9609235cce528e8b2280f5b2d18e688e611ac923346eb60c8ff2c5f9f77R998)
I don't see `username` and `preferred_username` in the `OpenIDClaims`

image

PR #344, opened 2022-12-05
add clusterAdmin property to cluster files

maorfr:
required for https://github.com/app-sre/qontract-reconcile/pull/3000

this field will be used by the integration to determine which clusters to act on.


chassing (approved):
LGTM

PR #348, opened 2022-12-06
cluster-1.yml#slack attribute

chassing:
* Add a dedicated `cluster-1` slack user group attribute to support clusters without GitHub authentication.
* Remove `slack-cluster-usergroups`

Tickets:
* slack-cluster-usergroups OIDC support ([APPSRE-6575](https://issues.redhat.com/browse/APPSRE-6575))
* Merge slack integrations ([APPSRE-6593](https://issues.redhat.com/browse/APPSRE-6593))

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -796,6 +796,11 @@ confs:
   - { name: url, type: string, isRequired: true }
   - { name: auth, type: VaultSecret_v1, isRequired: true}
 
+- name: ClusterSlack_v1
+  fields:
+  - { name: userGroup, type: string, isRequired: true }
+  - { name: channels, type: string, isList: true, isRequired: true}
+
 - name: Cluster_v1
   datafile: /openshift/cluster-1.yml
   fields:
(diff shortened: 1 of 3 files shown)


maorfr:
would it make sense to make this implicit?
the reason we manage these user groups is to notify of upgrades. upgrades depend on `ocm`. so if a cluster has an `ocm` section, (implicitly) manage the user group `<cluster_name>-cluster`?

in addition, perhaps we could allow disabling of `slack-usergroups` in a cluster file (`disable.integrations`)?

tl;dr in my view, this PR may not be required, if you accept this proposal.

chassing:
An implicit slack user group name is possible, and I can create a q-r method for that and reuse it in several integrations. 👍  But what about the channel? Currently, it comes from https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/dependencies/slack/coreos.yml#L62, which will be gone.

maorfr:
the channel is just a place holder because (i think) there must be at least one channel

chassing:
Yes, there must be one. Ok, then I'll reuse the `slack-workspace` config pattern.

PR #349, opened 2022-12-06
APPSRE-6644 Add channel to recommended version

janboll:
Along with this add initialVersion, that can be used by fleetmanagers to provision new clusters.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -899,7 +899,9 @@ confs:
 - name: OpenShiftClusterManagerRecommendedVersions_v1
   fields:
   - { name: recommendedVersion, type: string, isRequired: true }
+  - { name: initialVersion, type: string, isRequired: true }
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -899,7 +899,9 @@ confs:
 - name: OpenShiftClusterManagerRecommendedVersions_v1
   fields:
   - { name: recommendedVersion, type: string, isRequired: true }
+  - { name: initialVersion, type: string, isRequired: true }
```
patjlm, line 902: i am not sure of the wording here. The version should be the same as the `recommendedversion`. Only its format will be different, right? I think we mentioned `ocmVersion` or something like that, didn't we?

janboll: Funny, I remembered initial version, cause we use it in App-Interface ourselves.

PR #354, opened 2022-12-08
[pr-check] Shell GQL Logic Update

cubismod:
[APPSRE-6676](https://issues.redhat.com/browse/APPSRE-6676)

With the existing Makefile logic, it appears that Qontract Server is not run if `docker pull quay.io/app-sre/qontract-server` pulls a new version of the Qontract Server image. I'm not entirely sure why and troubleshoot using various different methods like splitting out to separate lines of the Makefile instruction but the issue persisted. Here, I'm splitting out the `docker pull` to a different Makefile instruction. Not 100% confident that this will fix the issue and it is difficult to test since the issue is so sporadic. However, I'll keep an eye on schemas checks and see if the behavior continues.


geoberle (approved):
good catch!

cubismod:
Sadly, the issue is persisting

PR #364, opened 2023-01-06
[pr-check] Move timeout into container

cubismod:
[APPSRE-6676](https://issues.redhat.com/browse/APPSRE-6676)

Move the timeout command to run within the container itself to avoid an issue where pulling the qontract-server container resulted in it not actually running.


cubismod:
The PR check produced the output I expected. It pulled the QS image and also ran the timeout command unlike before where it just pulled the image and never actually ran the gql check.

bhushanthakur93 (approved):
LGTM

PR #367, opened 2023-01-11
Support Cloudflare DNS Zones

OliviaHY:
### What:
* New schema for Cloudflare DNS zones
* Extract Cloudflare DNS record schema from terraform-resource-1
* New sharding strategy: per-cloudflare-zone
### Why:
 Be able to create and manage DNS zones in Cloudflare, then eventually migrate resource from Dyn since its DNS service will soon to be not supported.
### Validation:
Locally ran terraform-cloudflare-dns with new schemas was able to create records and zones based on dev-data 
Following query resulted in expected response:
```
{
  zones: cloudflare_dns_zone_v1 {
    zone
    account {
      name
      description
      providerVersion
      apiCredentials {
        path
        field
        format
        version
      }
      terraformStateAccount {
        name
        consoleUrl
        terraformUsername
        automationToken {
          path
          field
        }
      }
    }
    records {
      identifier
      name
      type
      ttl
      value
      priority
      data {
        algorithm
        protocol
        public_key
        digest_type
        digest
        key_tag
        flags
      }
    }    partial
    jump_start
    plan
  }
}
```
Data as following was added to local App Interface:
```
---
$schema: /cloudflare/dns-zone-1.yml

labels: {}

zone: devshiftfake.net

partial: true

account:
 $ref: /cloudflare/app-sre/account.yml
 
records:
 - name: cdn01
   identifier: cdn01
   type: CNAME
   ttl: 1
   value: visual-app-interface.devshift.net
   proxied: true
 - name: devshift_txt
   identifier: devshifttxt
   type: TXT
   ttl: 1
   value: printer=lpr5
   proxied: true```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3061,6 +3061,7 @@ confs:
   - { name: gabi_instances_v1, type: GabiInstance_v1, isList: true, datafileSchema: /app-sre/gabi-instance-1.yml }
   - { name: template_tests_v1, type: TemplateTest_v1, isList: true, datafileSchema: /app-interface/template-test-1.yml }
   - { name: dns_zone_v1, type: DnsZone_v1, isList: true, datafileSchema: /dependencies/dns-zone-1.yml }
+  - { name: cloudflare_dns_zone_v1, type: CloudflareDnsZone_v1, isList: true, datafileSchema: /cloudflare/dns-zone-1.yml }
   - { name: slack_workspaces_v1, type: SlackWorkspace_v1, isList: true, datafileSchema: /dependencies/slack-workspace-1.yml }
   - { name: ocp_release_mirror_v1, type: OcpReleaseMirror_v1, isList: true, datafileSchema: /dependencies/ocp-release-mirror-1.yml}
   - { name: slo_document_v1, type: SLODocument_v1, isList: true, datafileSchema: /app-sre/slo-document-1.yml}
@@ -3254,6 +3255,44 @@ confs:
   - { name: vpc, type: AWSVPC_v1 }
   - { name: origin, type: string, isRequired: true }
   - { name: records, type: DnsRecord_v1, isList: true }
+
+- name: CloudflareDnsRecordDataSettings_v1
+  fields:
+  - { name: algorithm, type: int }
+  - { name: key_tag, type: int }
+  - { name: flags, type: int }
+  - { name: protocol, type: int }
+  - { name: public_key, type: string }
+  - { name: digest_type, type: int }
+  - { name: digest, type: string }
+
+- name: CloudflareDnsRecord_v1
+  datafile: /cloudflare/dns-record-1.yml
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: identifier, type: string, isRequired: true, isUnique: true}
+  - { name: name, type: string, isRequired: true }
+  - { name: type, type: string, isRequired: true}
+  - { name: ttl, type: int, isRequired: true}
+  - { name: value, type: string }
+  - { name: data, type: CloudflareDnsRecordDataSettings_v1 }
+  - { name: proxied, type: boolean }
+  - { name: priority, type: int }
+
+
+- name: CloudflareDnsZone_v1
+  datafile: /cloudflare/dns-zone-1.yml
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: identifier, type: string, isRequired: true, isUnique: true}
+  - { name: account, type: CloudflareAccount_v1, isRequired: true }
+  - { name: zone, type: string, isRequired: true }
+  - { name: type, type: string }
+  - { name: plan, type: string }
+  - { name: records, type: CloudflareDnsRecord_v1, isList: true }
+  - { name: delete, type: boolean }
 
 - name: SLODocumentSLOSLOParameters_v1
   fields:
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: schemas/cloudflare/dns-record-1.yml [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+  name:
+    type: string
+  'type':
+    type: string
+    enum:
+    - A
+    - CNAME
+    - TXT
+    - NS
+    - DS # The record used to identify the DNSSEC signing key of a delegated zone
```
steveteahan, line 21: If we're supporting [DS and DNSKEY](https://github.com/cloudflare/terraform-provider-cloudflare/blob/02bf98d3154dec4fdaa5788d0a14c92733214ad6/internal/provider/schema_cloudflare_record.go#L176), will we need to also have a separate `data` field in the schema to map to the [Terraform provider schema field](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/record#data)?

OliviaHY: Good catch!

graphql-schemas/schema.yml
```diff
@@ -3217,6 +3218,30 @@ confs:
...
+  - { name: proxied, type: boolean }
+
+
+- name: CloudflareDnsZone_v1
+  datafile: /cloudflare/dns-zone-1.yml
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: zone, type: string, isRequired: true }
+  - { name: partial, type: boolean }
```
steveteahan, line 3239: Is there a use case for `partial` when this configuration is intended to manage the DNS zone?

> A partial (CNAME) setup allows you to use Cloudflare’s reverse proxy while maintaining your primary and authoritative DNS provider.

jfchevrette: It would be best to match the field with what terraform expects for better clarify and less need to translate in code

Ex for the other Cloudflare work I had a `type` field enum with `partial` and `full` values. See https://github.com/app-sre/qontract-schemas/blob/400758fc32492fddc7770b1be7cbf40da8f8e0d7/schemas/cloudflare/terraform-resource-1.yml#L52-L55

graphql-schemas/schema.yml
```diff
@@ -3217,6 +3218,30 @@ confs:
...
+  - { name: jump_start, type: boolean }
```
jfchevrette, line 3240: I'm not sure we have a real need or use case for this feature. I would remove it and leave it as default (it's optional in terraform)

schemas/cloudflare/dns-record-1.yml
OliviaHY: Add new CloudflareDnsRecordDataSettings_v1 schema to encapsulate all the data fields for DS and DNSKEY.

schemas/cloudflare/dns-zone-1.yml
```diff
@@ -0,0 +1,38 @@
...
+    - free
+    - enterprise
+  type:
+    enum:
+    - full
+    - partial
+  account:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef": "/cloudflare/account-1.yml"
+  records:
```
steveteahan, line 30: How are we enforcing the 1000 record limit? Does it make sense to do this in the schema? Configuration?

jfchevrette: In the schema we can use a `maxItems` to enforce a maximum. The good: it fails early at schema validation, the bad: the schema error is hard to understand (bad UX). I wonder if your implementation of json schema supports custom error messages [like this](https://github.com/networknt/json-schema-validator/blob/master/doc/cust-msg.md)?

Otherwise I'm in favor of an early check in the integration as we can generally make the UX better there

graphql-schemas/schema.yml
```diff
@@ -3218,6 +3219,41 @@ confs:
...
+- name: CloudflareDnsZone_v1
+  datafile: /cloudflare/dns-zone-1.yml
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: identifier, type: string, isRequired: true }
+  - { name: account, type: CloudflareAccount_v1, isRequired: true }
+  - { name: zone, type: string, isRequired: true }
+  - { name: type, type: string }
+  - { name: data, type: string }
```
jfchevrette, line 3252: What is this field for?

OliviaHY: This is a field for DS and DNSKEY records specifically, see https://github.com/app-sre/qontract-schemas/pull/367#discussion_r1068495774

jfchevrette: I see a `data` field in the DNS record schema for DS and DNSKEY records but this one here is in the zone schema which don't seem to exist in the provider?

schemas/cloudflare/dns-zone-1.yml
jfchevrette: There is no need for a custom error because the jsonschema package we use in `qontract-validator` already provides a relevant error. However due to some issue in either qontract-validator's error handling that nice error is not shown and instead a more generic schema error is shown instead

I'm trying to find a way to solve this but until then I think we can still leverate the `maxItems` schema validation keyword

graphql-schemas/schema.yml
OliviaHY: Ahh you are right, my mistake. Removing.

steveteahan (approved):

PR #371, opened 2023-01-17
add workerFleets to jenkins-instance

BumbleFeng:
https://issues.redhat.com/browse/APPSRE-6798

### DIFF
--- a/schemas/dependencies/jenkins-instance-1.yml
+++ b/schemas/dependencies/jenkins-instance-1.yml
@@ -50,6 +50,42 @@ properties:
       required:
       - name
       - keep_hours
+  workerFleets:
+    type: array
+    items:
+      type: object
+      additionalProperties: false
+      properties:
+        account:
+          "$ref": "/aws/tenant_accounts-1.yml#/properties/account"
+        identifier:
+          "$ref": "/common-1.json#/definitions/longIdentifier"
+          description: ASG used for this Jenkins cloud, need to match an ASG identifier.
+        namespace:
+          "$ref": "/common-1.json#/definitions/crossref"
+          "$schemaRef": "/openshift/namespace-1.yml"
+        credentialsId:
+          type: string
+          description: Jenkins credential name of ssh private key to be used for logging in to the remote host.
+        # following options comes form https://github.com/jenkinsci/ec2-fleet-plugin/blob/master/docs/
+        fsRoot:
+          type: string
+        labelString:
+          type: string
+        numExecutors:
+          type: integer
+        idleMinutes:
+          type: integer
+        minSpareSize:
+          type: integer
+        noDelayProvision:
+          type: boolean
+      required:
+      - account
+      - identifier
+      - namespace
+      - credentialsId
+      - labelString
 required:
 - "$schema"
 - labels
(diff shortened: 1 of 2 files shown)


schemas/dependencies/jenkins-instance-1.yml
```diff
@@ -50,6 +50,39 @@ properties:
       required:
       - name
       - keep_hours
+  workerFleets:
+    type: array
+    items:
+      type: object
+      additionalProperties: false
+      properties:
```
rporres, line 58: I would add a description field here stating how this options come from https://github.com/jenkinsci/ec2-fleet-plugin/blob/master/docs/

schemas/dependencies/jenkins-instance-1.yml
```diff
@@ -50,6 +50,39 @@ properties:
...
+          "$ref": "/common-1.json#/definitions/longIdentifier"
+        namespace:
+          "$ref": "/common-1.json#/definitions/crossref"
+          "$schemaRef": "/openshift/namespace-1.yml"
+        credentialsId:
+          type: string
+        fsRoot:
+          type: string
+        labelString:
+          type: string
```
rporres, line 71: I guess it's important to state here that this has to match the tags given in the `asg` configuration

rporres (commented):
a couple of documentation requests, looks good otherwise.

schemas/dependencies/jenkins-instance-1.yml
```diff
@@ -50,6 +50,39 @@ properties:
...
       - keep_hours
+  workerFleets:
+    type: array
+    items:
+      type: object
+      additionalProperties: false
+      properties:
+        account:
+          "$ref": "/aws/tenant_accounts-1.yml#/properties/account"
+        identifier:
```
rporres, line 61: This identifier needs to match an asg identifier, correct? If so, we should document it here.

schemas/dependencies/jenkins-instance-1.yml
```diff
@@ -50,6 +50,39 @@ properties:
...
       - name
       - keep_hours
+  workerFleets:
+    type: array
+    items:
+      type: object
+      additionalProperties: false
+      properties:
+        account:
+          "$ref": "/aws/tenant_accounts-1.yml#/properties/account"
```
rporres, line 60: Do we actually need this, or since we have to match the ASG it is redundant?

rporres: well, we can have two ASGs with the same name in different accounts, so this is pertinent.

schemas/dependencies/jenkins-instance-1.yml
```diff
@@ -50,6 +50,42 @@ properties:
...
+        identifier:
+          "$ref": "/common-1.json#/definitions/longIdentifier"
+          description: ASG used for this Jenkins cloud, need to match an ASG identifier.
+        namespace:
+          "$ref": "/common-1.json#/definitions/crossref"
+          "$schemaRef": "/openshift/namespace-1.yml"
+        credentialsId:
+          type: string
+          description: Jenkins credential name of ssh private key to be used for logging in to the remote host.
+        # following options comes form https://github.com/jenkinsci/ec2-fleet-plugin/blob/master/docs/
```
rporres, line 70:
```suggestion
        # following options come form https://github.com/jenkinsci/ec2-fleet-plugin/blob/master/docs/
```

rporres (approved):

PR #375, opened 2023-02-06
indicate an rds instance is used as a grafana datasource

maorfr:
this can be used for multiple purposes, such as audit, self-service, automation, and more.


geoberle:
so e.g. add such a flagged RDS instance to grafana automatically? or at least report on it if it is present there without being flagged?

maorfr:
absolutely. and perhaps more around compliance, and anything else we may need.

maorfr:
related to https://issues.redhat.com/browse/APPSRE-6791

pontg (approved):

geoberle (approved):
valuable for compliance!

PR #377, opened 2023-02-07
[glitchtip] get automation user email from vault secret

chassing:
Refactor `glitchtip-instance-1` and get the automation user email address from a vault secret instead of specifying it manually. The vault secret is used by the glitchtip application during the startup to create/reconcile the API users.
This avoids glitchtip instance misconfigurations.

Ticket: [APPSRE-6904](https://issues.redhat.com/browse/APPSRE-6904)


fishi0x01 (approved):
LGTM

PR #396, opened 2023-02-27
remove jira ref from code components

maorfr:
this has not been in use for a long time now, since the removal of the `gitlab-integrations` integration.


kwilczynski (approved):
@maorfr, looks good!

PR #397, opened 2023-02-28
Revert "remove jira ref from code components"

bhushanthakur93:
Reverts app-sre/qontract-schemas#396

This is still being used in `gitlab-projects` and `gitlab-labeler` integrations.


fishi0x01 (approved):
LGTM :+1:

PR #398, opened 2023-03-02
glitchtip-project-dsn cluster.disable property

chassing:
The `glitchtip-project-dsn` integration can be disabled per cluster


fishi0x01 (approved):
LGTM

PR #400, opened 2023-03-02
aws/asg-defaults-1: add support for `price-capacity-optimized`

croissanne:
This new strategy strikes a balance between cost and interruption risk.

The `spot_instance_pools` parameter is only available with the
    `lowest-price` strategy, so it needs to be set to 0 for other
    strategies.

https://aws.amazon.com/blogs/compute/introducing-price-capacity-optimized-allocation-strategy-for-ec2-spot-instances/

### DIFF
--- a/schemas/aws/asg-defaults-1.yml
+++ b/schemas/aws/asg-defaults-1.yml
@@ -40,11 +40,14 @@ properties:
         type: string
         enum:
         - lowest-price
+        - price-capacity-optimized
         - capacity-optimized
         - capacity-optimized-prioritized
       on_demand_base_capacity:
         type: integer
       on_demand_percentage_above_base_capacity:
+        type: integer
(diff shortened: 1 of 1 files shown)


croissanne:
Will open a PR in qontract-reconcile first, as we need to be able to specify `spot_instance_pools` when updating the strategy.

```
Error updating Auto Scaling Group: ValidationError: SpotInstancePools option is only available with the lowest-price allocation strategy.
```

Terraform seems to set this one by default, and we need to be able to unset it (to 0).

croissanne:
Looking at qontract-reconcile, @ondrejbudai remarked that all `instances_distribution` values are just copied over in bulk, including `spot_instance_pools`. So changing the strategy up should work as is.

BumbleFeng (approved):

PR #406, opened 2023-03-08
Increase the number of record in a Cloudflare DNS zone to 1500

OliviaHY:
Increase the number of record to 1500 to accommodate a large zone from Dyn


steveteahan (approved):
Approving with the assumption that we'll be adding soft limits to the integration to prevent teams from creating larger zones without our approval.

PR #416, opened 2023-03-20
APPSRE-7116: remove dyn-traffic-director schemas

jfchevrette:
Dyn is retiring their Managed DNS services, as such we are removing support for the Dyn Traffic Director integration

Related: removal of integration: https://github.com/app-sre/qontract-reconcile/pull/3342

Ref.: https://issues.redhat.com/browse/APPSRE-7116


fishi0x01 (approved):
Hopefully these schemas aren't used somewhere else in AI. Maybe update introspection.json in qr too, just in case?

jfchevrette:
> Hopefully these schemas aren't used somewhere else in AI. Maybe update introspection.json in qr too, just in case?

I double checked and they are not used elsewhere. I will re-generate introspection.json in https://github.com/app-sre/qontract-reconcile/pull/3342 once this one is merged. Thank you!

PR #419, opened 2023-03-22
Add cloudflare dns and account to self serviceable schema whitelist.

OliviaHY:
Add cloudflare dns and account to self serviceable schema whitelist.


bhushanthakur93 (approved):
LGTM

PR #420, opened 2023-03-23
add datafile reference to CloudflareAccount_V1 GQL type

geoberle:
add the jsonschema type to the CloudflareAccount_V1 GQL type. this is required for the `interfaceResolver.strategy = schema` configuration of `DatafileObject_V1`

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1140,6 +1140,7 @@ confs:
   - { name: addr_block, type: string }
 
 - name: CloudflareAccount_v1
+  datafile: /cloudflare/account-1.yml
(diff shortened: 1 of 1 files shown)


OliviaHY:
LGTM

OliviaHY (approved):
Thank you for making this change @geoberle I wonder we should add this to the qontract-schema so people understand the purpose of adding `datafile:` is partially to support self serviceability?

geoberle:
> Thank you for making this change @geoberle I wonder we should add this to the qontract-schema so people understand the purpose of adding `datafile:` is partially to support self serviceability?

we could add it here, wdyt? https://github.com/app-sre/qontract-server/blob/master/SCHEMA.md

PR #421, opened 2023-03-24
tf-cloudflare-resources: add tiered_cache cache_type setting

jfchevrette:
Adds ability to set [tiered_cache](https://registry.terraform.io/providers/cloudflare/cloudflare/3.32.0/docs/resources/tiered_cache) on a zone

Ref.: https://issues.redhat.com/browse/APPSRE-7363

### DIFF
--- a/schemas/cloudflare/terraform-resource-1.yml
+++ b/schemas/cloudflare/terraform-resource-1.yml
@@ -63,6 +63,17 @@ properties:
         type: boolean
       smart_routing:
         type: boolean
+  tiered_cache:
+    type: object
+    additionalProperties: false
+    properties:
+      cache_type:
+        type: string
+        enum:
+          - generic
+          - smart
+    required:
+      - cache_type
   cache_reserve:
     type: object
     additionalProperties: false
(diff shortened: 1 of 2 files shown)


schemas/cloudflare/terraform-resource-1.yml
```diff
@@ -63,6 +63,17 @@ properties:
...
         type: boolean
+  tiered_cache:
+    type: object
+    additionalProperties: false
+    properties:
+      cache_type:
+        type: string
+        enum:
+          - generic
+          - smart
```
steveteahan, line 74: nit: is `off` not necessary: https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/tiered_cache
?

I don't know a lot of use cases immediately, but maybe something when debugging?

steveteahan (approved):

schemas/cloudflare/terraform-resource-1.yml
jfchevrette: Their API docs and terraform docs don't match up. I went with what they have available in the web console which is the two options (generic and smart). From what I can understand from looking at the API and provider code, `off` is taken care of if we set `argo.tiered_caching` to `off`

PR #423, opened 2023-03-28
vault secret path must not start or end with a /

chassing:
To avoid `Replacing double-slashes ("//") in path with single slash ("/") to avoid Vault redirect response.` openshift-vault-secret warnings, enforce vault secret path without leading or ending /

Ticket: [APPSRE-7398](https://issues.redhat.com/browse/APPSRE-7398)


fishi0x01 (approved):
:tada:

PR #424, opened 2023-03-28
Adding soft limit for Cloudflare DNS zone

OliviaHY:
- Add max_records to zone
- Add default max records to App Interface settings


bhushanthakur93 (approved):
LGTM

PR #425, opened 2023-03-29
Terraform Repo Schema

cubismod:
[APPSRE-7061](https://issues.redhat.com/browse/APPSRE-7061)

New schema for a Terraform Repo which is not associated with a namespace but rather an AWS account. This is separate from a "Terraform Resource" which is handled by Terrascript and a separate integration. 

Corresponding design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/58341#note_5807974

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3354,6 +3354,7 @@ confs:
   - { name: pgp_reencrypt_settings_v1, type: PgpReencryptSettings_v1, isList: true, datafileSchema: /app-sre/pgp-reencrypt-settings-1.yml }
   - { name: skupper_network_v1, type: SkupperNetwork_v1, isList: true, datafileSchema: /dependencies/skupper-network-1.yml }
   - { name: cloudflare_account_role_v1, type: CloudflareAccountRole_v1, isList: true, datafileSchema: /cloudflare/account-role-1.yml }
+  - { name: terraform_repo_v1, type: TerraformRepo_v1, isList: true, datafileSchema: /aws/terraform-repo-1.yml }
 
 - name: GlitchtipInstance_v1
   fields:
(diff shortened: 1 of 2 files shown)


bkez322 (approved):
LGTM

graphql-schemas/schema.yml
```diff
@@ -1506,6 +1506,15 @@ confs:
   - { name: name, type: string, isRequired: true }
   - { name: text, type: string, isRequired: true }
 
+- name: TerraformRepo_v1
+  fields:
+  - { name: account, type: AWSAccount_v1, isRequired: true }
```
OliviaHY, line 1511: For my understanding, I saw it in the design doc that each repo will need to be able to target multiple AWS account, how come this is only tied to one?

cubismod: That's a good point. This was done to allow for bootstrapping of each terraform repo. A tenant could target multiple AWS accounts though by duplicating the terraform repo file and then targeting other AWS accounts. They can even re-use the existing Git repository and just change project paths. The only constraint is that each of terraform repos must have a unique name to avoid tfstate key file collisions.

OliviaHY: Thank you for the explanation. So from App Interface perspective, we can't really tell which AWS account a bootstrapped repo really targets, we'd go in the repo and look for `./stage-aws-account-name/main.tf
` etc?

cubismod: No, from App Interface you would be able to tell by looking at the terraform repo file, here's an example:

```yaml
---
$schema: /aws/terraform-repo-1.yml

account:
    $ref: /aws/app-int-example-01/account.yml

name: plan_a
repository: <my_repo>
ref: <my_ref>
projectPath: plan_a
```

Each repo would have a file like this in A-I so you could distinguish which repo is being applied to which account.

OliviaHY: Right but what if the repo is applying to more than one account?

PR #431, opened 2023-04-18
Refactor ALB rule conditions to support multiple condition types

jfchevrette:
This refactor allows (and adds) more condition types and paves the way to others, as needed

The schema maps to the terraform provider resource arguments and parameters so we can refer to the terraform docs

Before
```yaml
rules:
- condition:
    path: /foo/bar/*
  action:
```

After
```yaml
rules:
- condition:
  - type: path-pattern
    path_pattern:
    - /foo/bar/*
  # New condition types
  - type: host-header
    host_header:
    - foo.example.com
  - type: http-request-method
    http_request_method:
    - GET
    - HEAD
  - type: source-ip
    source_ip:
    - 10.11.12.13
```


graphql-schemas/schema.yml
```diff
@@ -1977,13 +1977,45 @@ confs:
...
+  interfaceResolve:
+    strategy: fieldMap
+    field: type
+    fieldMap:
+      host-header: NamespaceTerraformResourceALBConditionHostHeader_v1
+      http-request-method: NamespaceTerraformResourceALBConditionHTTPRequestMethod_v1
+      path-pattern: NamespaceTerraformResourceALBConditionPathPattern_v1
+      source-ip: NamespaceTerraformResourceALBConditionSourceIP_v1
+  fields:
+  - { name: type, type: string, isRequired: true }
```
steveteahan, line 1994: I'm sure there's a good reason for it, but I was just curious if `type` is redundant here?

```yaml
  - type: source-ip
    source_ip:
    - 10.11.12.13
```

I'm guessing there are difficulties with moving `source_ip` to being a field on `NamespaceTerraformResourceALBCondition_v1`?

It's not the biggest deal, just more curious than anything.

steveteahan (approved):

PR #438, opened 2023-05-03
add validate targets in app attribute to saas file

maorfr:
with such a flag set to `true`, we can put more trust in a saas file owner and grant them wider permissions, such as full management over `targets`.


patjlm (approved):

maorfr:
https://issues.redhat.com/browse/APPSRE-7587

PR #448, opened 2023-05-25
Add group label to Prometheus rule schema

saswatamcode:
This PR adds a `group` label to the Prometheus rule schema, as it is sometimes necessary to partition metrics according to API group, and query them in a rule.


hemslo (approved):
LGTM! Thanks!

PR #449, opened 2023-05-25
APPSRE-6899 extend sql query whitelist

fishi0x01:
We want to allow more queries to accommodate for database post-upgrade queries. Specifically allow:

- ANALYZE
- VACUUM
- REINDEX

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -5,7 +5,7 @@
   "definitions": {
     "query": {
       "type": "string",
-      "pattern": "(?i)^(?:select|explain(?!\\s+analyze(?!\\s+select)))(?:(?:\\/\\*.*\\*\\/)|(?:[\\\"\\'][^\\\"\\']*[\\\"\\'])|(?:--.*\\n)|[^;\\*])+;$"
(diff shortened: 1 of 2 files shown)


schemas/common-1.json
```diff
@@ -5,7 +5,7 @@
   "definitions": {
     "query": {
       "type": "string",
-      "pattern": "(?i)^(?:select|explain(?!\\s+analyze(?!\\s+select)))(?:(?:\\/\\*.*\\*\\/)|(?:[\\\"\\'][^\\\"\\']*[\\\"\\'])|(?:--.*\\n)|[^;\\*])+;$"
+      "pattern": "(?i)^(?:analyze|reindex|vacuum|select|explain(?!\\s+analyze(?!\\s+select)))(?:(?:\\/\\*.*\\*\\/)|(?:[\\\"\\'][^\\\"\\']*[\\\"\\'])|(?:--.*\\n)|[^;\\*])*;$"
```
rporres, line 8: Not a fan of that final `*` as that may lead to non-complete sql statements.

If you want to allow bare `ANALYZE;` you may want to contemplate that case separately...

rporres (approved):

schemas/common-1.json
kwilczynski, line 8: We don't want to include **OPTIMIZE** there to support MySQL? For example, Quay folks might want to run it following database upgrades.

PR #451, opened 2023-05-26
[Apache Managed Streaming for Kafka] MSK schemas

chassing:
Schemas to support AWS MSK

Ticket: [APPSRE-7534](https://issues.redhat.com/browse/APPSRE-7534)

### DIFF
--- a/schemas/aws/msk-defaults-1.yml
+++ b/schemas/aws/msk-defaults-1.yml
@@ -65,6 +65,7 @@ properties:
             - enabled_in_broker
     required:
     - prometheus
+  # we support logging to cloudwatch only
   logging_info:
     type: object
     additionalProperties: false
@@ -90,6 +91,7 @@ properties:
     type: object
     additionalProperties: false
     properties:
+      # tls authentication is not supported yet. Our tentants are using SASL at the moment.
       sasl:
         type: object
         additionalProperties: false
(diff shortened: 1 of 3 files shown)


maorfr (approved):

schemas/aws/msk-defaults-1.yml
```diff
@@ -0,0 +1,105 @@
...
+            additionalProperties: false
+            properties:
+              enabled_in_broker:
+                type: boolean
+            required:
+            - enabled_in_broker
+    required:
+    - prometheus
+  logging_info:
+    type: object
```
rporres, line 69: It would be nice to state that we consciously only support logging to cloudwatch.

schemas/aws/msk-defaults-1.yml
```diff
@@ -0,0 +1,105 @@
...
+            required:
+            - enabled
+            - retention_in_days
+    required:
+    - broker_logs
+  client_authentication:
+    type: object
+    additionalProperties: false
+    properties:
+      sasl:
```
rporres, line 93: why only `sasl` and not `tls` as supported by the tf module? It may be worth to add a comment somewhere, or in the commit message.

rporres (commented):
Looks fine. I would add some details of some of the decisions made in terms of what we support.

schemas/aws/msk-defaults-1.yml
chassing: I added a comment in the schemas, and I'll note this in the user documentation

schemas/aws/msk-defaults-1.yml
chassing: I left a comment in the schemas file.

rporres (approved):

PR #454, opened 2023-05-29
Add protocol and protocol_version to ALB provider (PROJQUAY-5569)

syed:
This is needed because we need to create a gRPC target group

### DIFF
--- a/schemas/aws/terraform-resource-2.yml
+++ b/schemas/aws/terraform-resource-2.yml
@@ -953,6 +953,21 @@ oneOf:
               type: string
           openshift_service:
             type: string
+          protocol:
+            type: string
+            enum:
+              - HTTP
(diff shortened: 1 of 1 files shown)


schemas/aws/terraform-resource-2.yml
```diff
@@ -961,6 +976,8 @@ oneOf:
           - ips
         - required:
           - openshift_service
+          - protocol
+          - protocol_version
```
patjlm, line 980: This makes those new fields required when `openshift_service` is set. Is that on purpose? Don't we want to rely on current defaults?

syed: @patjlm  remove the protocol and protocol_version to be rquired

jfchevrette (approved):

PR #455, opened 2023-05-30
add optional escalation policy to aws account and cluster

maorfr:
with an escalation policy referenced from an aws account or a cluster, we can use [dynamically generated per-resource alerts](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/sop/catch-all-alerts-routing.md) that will alert the tenant instead of AppSRE.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -881,6 +881,7 @@ confs:
   - { name: labels, type: json }
   - { name: name, type: string, isUnique: true, isSearchable: true, isRequired: true }
   - { name: description, type: string }
+  - { name: escalationPolicy, type: AppEscalationPolicy_v1 }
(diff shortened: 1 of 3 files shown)


patjlm:
Don't we have escalation policy from namespace->app already? Which scenario would require account/cluster link without namespace?

maorfr:
the scenario is someone looking for a cluster owner (that is not us) and not for a service owner.

geoberle:
i like this a lot, again a nice option to reduce alerts on us and hand them to the owners of clusters and accounts

geoberle (approved):

rporres:
I see the scenario of defining a cluster/account owner and I like the idea. I honestly don't see how this can be used to route catch-all alerts to tenants since that involve being able to associate alerts labels to apps (which is something I've been wanting to do for a while, but other stuff keeps coming).

PR #456, opened 2023-05-30
Schema addition for log retention

suzana-nesic:
Log groups are created with infinite log retention. All log groups should have a retention period of 3 months in order to keep storage costs in check. This change would create the following within an account file in app-interface:

```
---
$schema: /aws/account-1.yml

name: <aws-account-name>

cleanup:
- provider: cloudwatch
  regex: '<some-log-path>'
  retention_in_days: 90d
```

Note that this change is not affecting log groups managed by CLO.

Part of [APPSRE-7249](https://issues.redhat.com/browse/APPSRE-7249)

### DIFF
--- a/schemas/aws/cleanup-option-1.yml
+++ b/schemas/aws/cleanup-option-1.yml
@@ -23,7 +23,7 @@ properties:
     "$ref": "/common-1.json#/definitions/dhmsDuration"
     description: AMI age from which we can consider deletion.
   retention_in_days:
-    type: string
+    type: integer
     description: Cloudwatch log retention
 oneOf:
 - properties:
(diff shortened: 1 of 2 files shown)


schemas/aws/cleanup-option-1.yml
```diff
@@ -39,5 +42,19 @@ oneOf:
...
+- properties:
+    provider:
+      type: string
+      description: Set CloudWatch log retention
+      enum:
+      - cloudwatch
+    regex:
+      type: string
+    retention_in_days:
+      type: string
```
BumbleFeng, line 54: Since `retention_in_days` already indicates the measuring unit. I suggest removing the suffix 'd' and changing the type to integer. Besides, we could set an enum for the values, for reference https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html#API_PutRetentionPolicy_RequestSyntax

BumbleFeng (approved):

PR #457, opened 2023-06-01
Add new sharding per-aws-tgw-account

hemslo:
New sharding strategy, very similar to `per-aws-account`, but will filter only accounts with transit gateway.

[APPSRE-7624](https://issues.redhat.com/browse/APPSRE-7624)


hemslo:
Close this as there is a new solution

PR #464, opened 2023-06-07
Remove target_clusters as it is not longer used

rporres:
part of APPSRE-7549


rporres:
This is a breaking change: app-interface has already been cleaned-up and there are no reported prometheus tests in FedRAMP.

PR #467, opened 2023-06-15
[terraform msk] multi user support

chassing:
Support the creation of multiple MSK SASL/SCRAM users.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1731,7 +1731,7 @@ confs:
   - { name: output_resource_name, type: string }
   - { name: output_format, type: NamespaceTerraformResourceOutputFormat_v1, isInterface: true }
   - { name: annotations, type: json }
-  - { name: secret, type: VaultSecret_v1 }
+  - { name: users, type: SaasSecretParameters_v1, isList: true }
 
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -1731,7 +1731,7 @@ confs:
   - { name: output_resource_name, type: string }
   - { name: output_format, type: NamespaceTerraformResourceOutputFormat_v1, isInterface: true }
   - { name: annotations, type: json }
-  - { name: secret, type: VaultSecret_v1 }
+  - { name: users, type: SaasSecretParameters_v1, isList: true }
```
hemslo, line 1734: Is this schema a WIP that has no usages yet?
if there are usages, this is a breaking change (remove `secret`), then we need to

* add `users`
* switch all usages from `secret` to `users`
* delete `secret`

chassing: This schema is already in use for one [POC MSK cluster](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/assisted-installer/namespaces/assisted-installer-stage.yml#L158) and I'll adapt this config with the QR promotion

hemslo (approved):
LGTM!

PR #469, opened 2023-06-16
[terraform-msk] make users name attribute an identifier

chassing:
The `users.name` attribute is used as an AWS resource identifier. Make this attribute an identifier in the schemas and limit it to 64 chars.


kwilczynski (approved):
@chassing, looks good!

PR #471, opened 2023-06-19
Enable terraform-tgw-attachments to be disabled

hemslo:
Add `terraform-tgw-attachments` to the list so it can be disabled


kwilczynski (approved):
@hemslo, looks good!

PR #472, opened 2023-06-19
Limit the name length of the the pipelines

rporres:
part of [APPSRE-6031](https://issues.redhat.com/browse/APPSRE-6031)

### DIFF
--- a/schemas/app-sre/tekton-pipeline-template-1.yml
+++ b/schemas/app-sre/tekton-pipeline-template-1.yml
@@ -1,0 +1,59 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+description: |
+  Properties of tekton pipeline templates deployed in tekton pipelines providers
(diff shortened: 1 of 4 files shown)


schemas/app-sre/tekton-pipeline-template-1.yml
```diff
@@ -0,0 +1,59 @@
...
+    - onePerNamespace
+  path:
+    type: string
+  variables:
+    type: object
+oneOf:
+- properties:
+    name:
+      type: string
+      maxLength: 61
```
hemslo, line 30: Not sure why the max length is 61, in description we know the hard limit is 63, what are the 2 characters between them?

hemslo (approved):
LGTM!

schemas/app-sre/tekton-pipeline-template-1.yml
rporres: The prefix. Until it's removed, this needs to be 61.

PR #473, opened 2023-06-20
aws/asg-defaults: allow specifying instance_requirements

croissanne:
Adds the ability to specify vcpu, memory, and excluded instance types. While instance_requirements supports more parameters, expose just the ones needed for basic usage at first.

app-sre/qontract-reconcile#3595

### DIFF
--- a/schemas/aws/asg-defaults-1.yml
+++ b/schemas/aws/asg-defaults-1.yml
@@ -89,6 +89,27 @@ properties:
               - st1
             encrypted:
               type: boolean
+  instance_requirements:
+    type: object
+    additionalProperties: false
+    properties:
+      vcpu_count:
+        type: object
+        additionalProperties: false
+        properties:
+          min:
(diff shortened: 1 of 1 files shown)


schemas/aws/asg-defaults-1.yml
```diff
@@ -89,6 +89,27 @@ properties:
...
+          min:
+            type: integer
+          max:
+            type: integer
+      memory_mib:
+        min:
+          type: integer
+        max:
+          type: integer
+      excluded_intance_types:
```
rporres, line 109:
```suggestion
      excluded_instance_types:
```

rporres (commented):
Looks good (just a small typo to fix). I would add somewhere (in the commit message most likely) that the `instance_requirements` object has more options and we have only added here those that we need for the moment.

rporres (approved):

PR #485, opened 2023-07-28
Add stable as a supported branch name for SaaS template

Victoremepunto:
https://issues.redhat.com/browse/RHCLOUD-27381

This should add support for `stable` as a valid branch name to deploy console.redhat.com Frontend applications from stable branches

### DIFF
--- a/schemas/app-sre/saas-file-target-1.yml
+++ b/schemas/app-sre/saas-file-target-1.yml
@@ -37,7 +37,7 @@ properties:
       - include
   ref:
     type: string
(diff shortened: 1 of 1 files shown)


bennyturns:
Note: This is not just for insights, this is for every app that runs on console.redhat.com.

Victoremepunto:
> Note: This is not just for insights, this is for every app that runs on console.redhat.com.

Amended the description

maorfr (approved):

maorfr:
this is different from #120 since `release` indicates production, and `stable` not as much IMO

PR #492, opened 2023-08-09
APPSRE-8144 add dynatrace environment

fishi0x01:
This introduces the dynatrace environment as a dependency. That dependency can be used by clusters to specify where to send logs to. Additionally, namespaces have a flag to turn on/off dynatrace log ingestion.

Further, this schema can be easily extended to inhibit token information.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -572,6 +572,14 @@ confs:
   - { name: icon_emoji, type: string, isRequired: true }
   - { name: username, type: string, isRequired: true }
 
+- name: DynatraceEnvironment_v1
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: name, type: string, isRequired: true, isUnique: true }
+  - { name: description, type: string, isRequired: true }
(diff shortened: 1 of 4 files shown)


OliviaHY:
LGTM!

OliviaHY (approved):

maorfr:
some more information may be required here to serve https://github.com/app-sre/qontract-reconcile/pull/3734, such as a token from vault for each tenant.

fishi0x01:
> some more information may be required here to serve https://github.com/app-sre/qontract-reconcile/pull/3734, such as a token from vault for each tenant.

This is a first step towards that direction. Token provisioning schema will be addressed in a follow up PR. This PR strictly adds the stuff from design doc, w/o token provisioning - the design doc just roughly mentions token provisioning as a comment for future extension possibilities.

fishi0x01:
To avoid any misunderstanding: I will open a follow-up PR with the token included shortly.

The reason I didnt add it here is bc this schema change already entails a bigger MR in app-interface https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/76808 -> I wanted to keep the scope as small as possible and close to the design doc to get this done.

Now that this seems to be working, I will open follow-up with token :)

PR #493, opened 2023-08-09
Add issueType to jira board schema

ralphbean:
This permits a team to specify a new issue type for jiralert to use.

See downstream: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/76908

Issue https://issues.redhat.com/browse/RHTAPSRE-129

### DIFF
--- a/schemas/dependencies/jira-board-1.yml
+++ b/schemas/dependencies/jira-board-1.yml
@@ -31,6 +31,8 @@ properties:
         type: string
     required:
     - workspace
+  issueType:
+    type: string
   issueResolveState:
     type: string
   issueReopenState:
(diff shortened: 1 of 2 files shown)


schemas/dependencies/jira-board-1.yml
```diff
@@ -31,6 +31,13 @@ properties:
         type: string
     required:
     - workspace
+  issueType:
+    type: string
+    enum:
+    - Task
+    - Bug
+    - Ticket
+    - Incident
```
geoberle, line 40: i understand the idea of having a sanity check for issue types at the schema level. afaik issue types are jira instance and project specific (e.g. ticket and incident is not something i can use in our team project). hence we can't rely on this to make sure an alert will be created successfully on a specific board. 

i suggest to remove to enum

ralphbean: Makes sense. Will drop it.

ralphbean: Done.

geoberle: thanks

geoberle (approved):
LGTM. thanks for your contribution

PR #496, opened 2023-08-11
make change-type context.selector optional

patjlm:
It will default to `$.path` thanks to https://github.com/app-sre/qontract-reconcile/pull/3663

cc @geoberle


patjlm:
closing as
> right now it is still required and the code would fail if it is not there

https://github.com/app-sre/qontract-reconcile/pull/3663#issuecomment-1674242164

PR #499, opened 2023-08-15
introduce lean namespace

maorfr:
in some cases a lean definition of a namespace in a cluster with an automation token and some context (env) is all that is needed.

this is a proposal of such an implementation that works for integrations-manager with 0 code changes.

### DIFF
--- a/schemas/app-sre/integration-1.yml
+++ b/schemas/app-sre/integration-1.yml
@@ -76,7 +76,7 @@ properties:
                 type: string
                 enum:
                 - /openshift/namespace-1.yml
-                - /openshift/namespace-reference-1.yml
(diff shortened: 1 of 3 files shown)


maorfr:
example implementation: https://gitlab.cee.redhat.com/service/sre-capabilities/-/merge_requests/6

rporres:
> future thoughts: lean namespace would be a reference to the real definition in another app-interface instance.

Yeah, I like the idea. I would create a separate schema for that, so that we can keep the idea of lean namespace as the minimum building block for things like running integrations manager.

rporres (approved):

patjlm (approved):
Will be much better!

geoberle:
❤️

PR #502, opened 2023-08-21
remove enum for rds intsance class overrides

maorfr:
we don't need to enforce, nor to maintain a list of instance classes.


rporres:
As long as terraform-resources fails graciously in dry-run and not in apply, I'm fine with removing this enum.

patjlm:
I think we should keep this list and regularly reduce the number of non-graviton instances classes we make available

PR #515, opened 2023-09-14
__identifier field on crossrefs to enable validation of context uniqeness in lists of crossref objects

geoberle:
the `__identifier` field is a field managed by qontract-validator when an object defines one or many `isContextUnique` fields. it holds a hash which is then used to figure out if an object is really unique within the context.

to enable the context uniqueness check also on lists of crossrefs, the crossref jsonschema type needs to declare it so that qontract-bundler can add it and qontract-validator can still validate the schema.

completements https://github.com/app-sre/qontract-validator/pull/58
part of https://issues.redhat.com/browse/APPSRE-8308

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -141,6 +141,10 @@
         "$ref": {
           "type": "string",
           "format": "uri-reference"
+        },
+        "__identifier": {
+          "type": "string",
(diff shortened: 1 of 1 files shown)


chassing (approved):
LGTM

PR #516, opened 2023-09-14
Allow branch name `stage`

porridge:
We'd like to be able to have our:
- `integration` environment to follow the `main` branch
- `stage` environment to follow the `stage` branch
- `production` environment to be pinned to a commit

Currently this [fails validation](https://ci.int.devshift.net/job/service-app-interface-gl-pr-check/281770/artifact/temp/reports/index.html).

### DIFF
--- a/schemas/app-sre/saas-file-target-1.yml
+++ b/schemas/app-sre/saas-file-target-1.yml
@@ -37,7 +37,7 @@ properties:
       - include
   ref:
     type: string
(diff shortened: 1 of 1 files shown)


maorfr:
this is the preferred pattern: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/continuous-delivery-in-app-interface.md#progressive-rollouts-and-gated-promotions

porridge:
We'll stick to a hash ref for now.

PR #517, opened 2023-09-20
extract managedResourceNames

maorfr:
part of https://issues.redhat.com/browse/APPSRE-4355

also adds the important `additionalProperties: false`

### DIFF
--- a/schemas/openshift/managed-resource-name-1.yml
+++ b/schemas/openshift/managed-resource-name-1.yml
@@ -1,0 +1,23 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /openshift/managed-resource-name-1.yml
+  resource:
+    type: string
+  resourceNames:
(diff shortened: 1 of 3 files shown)


schemas/openshift/managed-resource-name-1.yml
```diff
@@ -0,0 +1,23 @@
...
+    type: string
+  resourceNames:
+    type: array
+    items:
+      type: string
+    minItems: 1
+
+required:
+- resource
+- resourceNames
```
patjlm, line 23: is `resourceNames` really required?

maorfr: what would we do without it?

patjlm: only limit the kind? I think we do that most of the time. Don't we?

maorfr: this is `managedResourceNames`, not `managedResourceTypes`

patjlm: 🤦

patjlm (approved):

PR #519, opened 2023-09-22
add soakDays for saas promotions

maorfr:
the idea would be to store version history continuously and only promote once soakDays passed

this concept is similar to the one of AUS, and may bring a standard to our different promotion processes.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2645,7 +2645,7 @@ confs:
   - { name: auto, type: boolean }
   - { name: publish, type: string, isList: true }
   - { name: subscribe, type: string, isList: true }
-  - { name: soakDays, type: integer }
+  - { name: soakDays, type: int }
   - { name: promotion_data, type: PromotionData_v1, isList: true }
 
 - name: SaasSecretParameters_v1
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2645,6 +2645,7 @@ confs:
   - { name: auto, type: boolean }
   - { name: publish, type: string, isList: true }
   - { name: subscribe, type: string, isList: true }
+  - { name: soakDays, type: integer }
   - { name: promotion_data, type: PromotionData_v1, isList: true }
 
 - name: SaasSecretParameters_v1
--- a/schemas/app-sre/saas-file-target-1.yml
+++ b/schemas/app-sre/saas-file-target-1.yml
@@ -54,6 +54,9 @@ properties:
         items:
           type: string
         minItems: 1
+      soakDays:
+        type: integer
+        description: number of days to wait for promotion
       promotion_data:
         type: array
         items:


patjlm (approved):

graphql-schemas/schema.yml
```diff
@@ -2645,6 +2645,7 @@ confs:
   - { name: auto, type: boolean }
   - { name: publish, type: string, isList: true }
   - { name: subscribe, type: string, isList: true }
+  - { name: soakDays, type: integer }
```
patjlm, line 2648:
```suggestion
  - { name: soakDays, type: int }
```

maorfr: https://github.com/app-sre/qontract-schemas/pull/519/commits/378ad5d1fdf3094da6a12d6253f26afd932d248c

jmelis:
Thanks a lot, this is fantastic! Let me tell you about our use case:

```mermaid
graph LR
    classDef gate fill:#fff3d1,stroke:#ffe59e;

    envA-deploy --> envA-gate:::gate
    envB-deploy --> envB-gate:::gate
    envB-gate:::gate --> soak
    envA-gate:::gate --> soak
```

So two things to note:
- The `soak` happens after one or more jobs finish. In the diagram above it's 2.
- The `soak` does not deploy anything, it doesn't touch the clusters. Right now we (mostly @ritmun) are  applying no-op template which is basically empty.

My initial thought for an RFE was therefore to:
- support empty targets in the saasfile, where you don't do anything, just run the promotion stuff
- what you are doing in this MR which is basically that you won't trigger the PR to move on to the next stage only after the soak time has passed.

So thanks a lot, I think this is exactly what we need :)

One more thing, it would be very nice to be able to "observe" the progress of the soak in the S3 Publisher data, is this possible? That way in the promotion DAG we can show this information ("soaking for 2 out of 3 days...")

jmelis:
Also, may I ask why this didn't move?

patjlm:
Why empty targets?
Do we all agree that if an other version is applied, the previous stops soaking? Can it still be used in some soakdays accumulation for next rollouts though (as we do for clusters)? I am not sure it's relevant in this case.
I have the impression this could "just" translate into storing the deployment timestamp into the publisher json, and really trigger a subscriber only after this soakday setting has passed (accumulating subscribers? or a different rule there?)

fishi0x01:
SAPM will add a `schedule` attribute to every MR. A `schedule` thats in the future will not have a `auto-promotion` label set. Once a MR passed the `schedule`, it will receive `auto-merge`.

Advantages:

- the tenant has full visibility on planned auto-promotions
- the tenant can manually lgtm a scheduled promotion prior to schedule
- the tenant can manually close a scheduled promotion (it wont be re-opened once commit sha in channel changed - i.e., the tenant can cancel old scheduled promotions)
- we dont need s3 state -> everything is persisted in MRs

maorfr:
https://issues.redhat.com/browse/APPSRE-10117

fishi0x01:
For transparency: we will align closer with AUS and not extend the MR based state.

PR #522, opened 2023-09-25
Cleanup cluster schema for default machine pool handling

hemslo:
`nodes`, `instance_type` and `autoscale` are no long used after https://github.com/app-sre/qontract-reconcile/pull/3825.

Also set `machinePools` to have at least 1 item.

[APPSRE-8258](https://issues.redhat.com/browse/APPSRE-8258)

### DIFF
--- a/schemas/openshift/cluster-1.yml
+++ b/schemas/openshift/cluster-1.yml
@@ -146,10 +146,6 @@ properties:
         type: string
       multi_az:
         type: boolean
-      nodes:
-        type: integer
-      instance_type:
-        type: string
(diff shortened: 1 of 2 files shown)


chassing (approved):

schemas/openshift/cluster-1.yml
```diff
@@ -257,6 +240,7 @@ properties:
         type: string
   machinePools:
     type: array
+    minItems: 1
```
janboll, line 243: I think you also want to make it required?

PR #523, opened 2023-09-27
fix some yaml lint attributes

maorfr:


patjlm (approved):

schemas/cloudflare/terraform-resource-1.yml
```diff
@@ -156,8 +156,6 @@ properties:
     type: string
   kind:
     type: string
-  name:
-    type: string
```
patjlm, line 160: why is that removed?

maorfr: https://github.com/app-sre/qontract-schemas/pull/523/commits/dfa00c23074eb65c5fdf1450059beac9596a4b30

patjlm: Ah I see, duplicate...

PR #529, opened 2023-10-04
[glitchtip] labels are not required

chassing:
To ease the glitchtip configuration, mark the `labels` attributes as non-required.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3557,6 +3557,7 @@ confs:
   fields:
   - { name: path, type: string, isRequired: true }
   - { name: schema, type: string, isRequired: true }
+  - { name: labels, type: json }
   - { name: name, type: string, isRequired: true, isUnique: true }
   - { name: projectId, type: string }
   - { name: description, type: string, isRequired: true }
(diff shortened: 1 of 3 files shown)


hemslo (approved):
LGTM!
How about `labels` in https://github.com/app-sre/qontract-schemas/blob/37d8d99c302452c7bbf399cefe773588eafed39e/schemas/dependencies/glitchtip-instance-1.yml#L27C3-L27C9

chassing:
> How about `labels` in https://github.com/app-sre/qontract-schemas/blob/37d8d99c302452c7bbf399cefe773588eafed39e/schemas/dependencies/glitchtip-instance-1.yml#L27C3-L27C9

I want to ease the tenants' configuration, especially for sre-capabilities. A tenant doesn't need to configure a glitchtip instance, so the `labels` attribute doesn't matter there :)

rporres:
Looks fine since no one is setting them anyway... What was the intended use?

chassing:
> Looks fine since no one is setting them anyway... What was the intended use?

No intended use at all, for me, `labels` is just a standard field.

rporres (approved):

PR #530, opened 2023-10-05
make serviceOwners optional

BumbleFeng:


geoberle:
hi @BumbleFeng 
can you link a ticket or elaborate on the reason for this change?

PR #537, opened 2023-10-15
disable integrations for an OCM organizations

geoberle:
disabling integrations per OCM organization (similar to AWS accounts and clusters)

part of https://issues.redhat.com/browse/APPSRE-8436


janboll (approved):

geoberle:
close in favour of #541

PR #540, opened 2023-10-19
/aws/parameter-group-1.yml schema

chassing:
To enable self-service on parameter groups, the parameter groups must have a schema.

This new `/aws/parameter-group-1.yml` schema supports MySQL, Postgres, and Redis (Elasticache).

Ticket: https://issues.redhat.com/browse/APPSRE-8310

### DIFF
--- a/schemas/aws/parameter-group-1.yml
+++ b/schemas/aws/parameter-group-1.yml
@@ -1,0 +1,54 @@
+---
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /aws/parameter-group-1.yml
+  name:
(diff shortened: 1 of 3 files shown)


hemslo (approved):
LGTM!

schemas/aws/parameter-group-1.yml
```diff
@@ -0,0 +1,54 @@
...
+          type: [string, integer, number, boolean]
+        apply_method:
+          type: string
+          enum:
+          - immediate
+          - pending-reboot
+      required:
+      - name
+      - value
+required:
```
hemslo, line 50: Do you also want to mark `name` and `identifier` as required?

chassing: Both attributes are not mandatory.

PR #546, opened 2023-11-06
declare membership sources in a role

geoberle:
We should be able to run integrations that rely on roles, users and role membership in the sre-capabilities repo with as little change as possible, while not being forced to manage a whole set of users and role memberships.

The central concept for grouping and permissions will still be `/access/role-1.yml` but membership will be pluggable. A configuration section memberSource can be used to bring in user and role membership information from different sources, defined via a schema `/access/membership-provider-1.yml`.

A membership provider can be any system that provides groups and users (provider pattern). User information needs to include some sort of username that can be mapped to org_username. This schema change introduces the memebership provider mechanism and offers app-interface as a first source for members.

```yaml
$schema: /access/membership-provider-1.yml

name: <provider-name>
source:
  provider: app-interface
  url: <url to qontract server of an a-i instance>
  username:
    $ref: <vault secret>
  password:
    $ref: <vault secret>
```

Then this provider can be used as follows

```yaml
---
$schema: /access/role-1.yml
...
memberSources:
- provider:
    $ref: <the provider>
  group: <a role name from the other app-interface>
```

APPSRE-8450

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3107,6 +3107,7 @@ confs:
   - { name: self_service, type: SelfServiceConfig_v1, isList: true }
   - { name: cloudflare_access, type: CloudflareAccountRole_v1, isList: true }
   - { name: ldapGroup, type: string }
+  - { name: memberSource, type: RoleMembershipSource_V1 }
   - name: users
     type: User_v1
     isList: true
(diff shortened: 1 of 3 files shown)


graphql-schemas/schema.yml
```diff
@@ -3107,6 +3107,7 @@ confs:
   - { name: self_service, type: SelfServiceConfig_v1, isList: true }
   - { name: cloudflare_access, type: CloudflareAccountRole_v1, isList: true }
   - { name: ldapGroup, type: string }
+  - { name: memberSource, type: RoleMembershipSource_V1 }
```
patjlm, line 3110: should that be a list?

geoberle: i kept it a single one for simplicity. 1-1 relationships are easier to understand
do you think we should have more than one?

geoberle: i made it a list

patjlm (approved):

PR #548, opened 2023-11-08
rename from targetNamespace to SLONamespace

jmelis:
`targetNamespace` is confusing, it has been renamed to `SLONamespace` and it now includes a description explaining its purpose.

### DIFF
--- a/schemas/app-sre/slo-document-1.yml
+++ b/schemas/app-sre/slo-document-1.yml
@@ -28,13 +28,15 @@ properties:
           "$schemaRef": "/openshift/namespace-1.yml"
         SLONamespace:
           description: |
-            This field is used to determine in which namespace to generate the PrometheusRules
-            that contain the dynamic alerts generated for the SLO, that are being used in the
-            slo-to-status-board flow:
+            This field is used to determine in which namespace PrometheusRule
+            object that contain the alerts generated from the expressions of
+            this SLO document is going to be created. Those alerts are being
+            used in the slo-to-status-board flow:
             https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/status-board/statusboard-alertmanager-receiver.md
 
-            If omitted, the rules will be generated in the Observability namespace of the cluster
-            to which `namespace` belongs to.
+            If omitted, the rules will be generated in the namespace
+            observabilityNamespace of the cluster to which the `namespace` of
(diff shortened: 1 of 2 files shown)


schemas/app-sre/slo-document-1.yml
```diff
@@ -26,7 +26,15 @@ properties:
...
-        targetNamespace:
+        SLONamespace:
+          description: |
+            This field is used to determine in which namespace to generate the PrometheusRules
+            that contain the dynamic alerts generated for the SLO, that are being used in the
+            slo-to-status-board flow:
+            https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/status-board/statusboard-alertmanager-receiver.md
+
+            If omitted, the rules will be generated in the Observability namespace of the cluster
+            to which `namespace` belongs to.
```
rporres, line 37:
```suggestion
            If omitted, the rules will be generated in the namespace observabilityNamespace of the cluster
            to which the `namespace` of this SLO document belongs to.
```

schemas/app-sre/slo-document-1.yml
```diff
@@ -26,7 +26,15 @@ properties:
         namespace:
           "$ref": "/common-1.json#/definitions/crossref"
           "$schemaRef": "/openshift/namespace-1.yml"
-        targetNamespace:
+        SLONamespace:
+          description: |
+            This field is used to determine in which namespace to generate the PrometheusRules
+            that contain the dynamic alerts generated for the SLO, that are being used in the
+            slo-to-status-board flow:
```
rporres, line 33: What about something like?

```
This field is used to determine in which namespace PrometheusRule object
that contain the alerts generated from the expressions of this SLO document
is going to be created. Those alerts are being used in the slo-to-status-board 
flow:
```

rporres: "is going to be created" can be is located. The key point is to avoid the feeling that something will automagically create the PrometheusRule

schemas/app-sre/slo-document-1.yml
jmelis: applied

schemas/app-sre/slo-document-1.yml
jmelis: 👍 applied

rporres (approved):

PR #551, opened 2023-11-17
[APPSRE-8492] removing placeholder change for ci.ext vault upgrade validations - after upgrade

suzana-nesic:
Part of [APPSRE-8492](https://issues.redhat.com/browse/APPSRE-8492)


bkez322 (approved):
Absolutely incredible great work.

PR #552, opened 2023-11-24
controlplane role is optional on HCP

geoberle:
HCP clusters don't have controlplane account roles

APPSRE-8561


hemslo (approved):
LGTM! It would be good to mark other fields as required in json schemas, it's strange that this required only set on graphql schema.

It would be good to update test files too https://github.com/app-sre/qontract-reconcile/blob/e7d92c397b4189308dc84d66639d15685f7944a0/reconcile/test/test_ocm_clusters.py#L262

geoberle:
> LGTM! It would be good to mark other fields as required in json schemas, it's strange that this required only set on graphql schema.

fixed. good catch. this inconsistency could easily cause problems during query time

PR #559, opened 2023-12-11
Adding region and environment information for Dynatrace environments.

OliviaHY:
### Why
OSDFM need to be able to generate configmap about the mapping of DT regions and HCP AWS regions.

### Validation:
I have run qontract server with this branch locally.

### DIFF
--- a/schemas/dependencies/dynatrace-environment-1.yml
+++ b/schemas/dependencies/dynatrace-environment-1.yml
@@ -14,11 +14,12 @@ properties:
   description:
     type: string
   environmentUrl:
-    description: the URL to a specific Dynatrace entity called environment, which is also known as tenant.
+    description: The URL to a specific Dynatrace entity called environment, which is also known as tenant.
     type: string
   bootstrapToken:
     "$ref": "/common-1.json#/definitions/vaultSecret"
   environment:
+    description: Which of integration, stage or production are the data coming from.
     type: string
     enum:
     - integration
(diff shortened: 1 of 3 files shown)


schemas/dependencies/dynatrace-environment-1.yml
```diff
@@ -14,9 +14,27 @@ properties:
...
+    type: string
+    enum:
+    - us-east-virginia
+    - eu-west-ireland
+    - ap-southeast-singapore
+  awsRegion:
+    description: a list of HCP AWS regions
+    type: array
+    items:
+      type: string
```
fishi0x01, line 37: Should these new fields map to a gql type too?

schemas/dependencies/dynatrace-environment-1.yml
```diff
@@ -14,9 +14,27 @@ properties:
   description:
     type: string
   environmentUrl:
+    description: the URL to a specific Dynatrace entity called environment, which is also known as tenant.
     type: string
   bootstrapToken:
     "$ref": "/common-1.json#/definitions/vaultSecret"
+  environment:
```
fishi0x01, line 21: Not super happy with the name, since we already use `environmentUrl` above which is sth completely different. Maybe we can add a description here that this purely targets different stages?

fishi0x01 (approved):

schemas/dependencies/dynatrace-environment-1.yml
OliviaHY: You mean `schemas/aws/regions-1.yml`?

fishi0x01: My understanding is we will need these as gql types too, e.g., [here](https://github.com/app-sre/qontract-schemas/blob/fda16f3a50edb8fe1e4345cbe0a8ec15f66a1fc0/graphql-schemas/schema.yml#L586)?

OliviaHY: Ah yes.

PR #562, opened 2023-12-14
[app-interface-settings] add jiralert settings

chassing:
Make some hardcoded Jiralert settings customizable. 

Ticket: [APPSRE-8620](https://issues.redhat.com/browse/APPSRE-8620)

### DIFF
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -198,6 +198,7 @@ properties:
   jiralert:
     type: object
     additionalProperties: false
+    # we can't use enum here because the values are highly customisable per jira project/instance
     properties:
       defaultIssueType:
         type: string
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -195,6 +195,20 @@ properties:
     required:
     - contactList
     - credentials
+  jiralert:
+    type: object
+    additionalProperties: false
+    properties:
+      defaultIssueType:
+        type: string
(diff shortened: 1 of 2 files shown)


schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -195,6 +195,20 @@ properties:
...
+  jiralert:
+    type: object
+    additionalProperties: false
+    properties:
+      defaultIssueType:
+        type: string
+      defaultPriority:
+        type: string
+      defaultReopenState:
+        type: string
```
rporres, line 207: Can we have enums here? Most likely jira will also need them

chassing: This was my first thought too, but I think it's better to keep plain `string`s because there aren't any defaults available in every Jira instance, and all of them can be very custom. e.g. the reopen (open) state can be something like `needs planning`.

rporres: I see... this is certainly going to bite us at some point, but what can we do.

Since this thought is going to certainly rise again to anyone reading this schema, a comment in the file or in the git message will certainly help to understand why there are no enums here :)

rporres (approved):

schemas/app-interface/app-interface-settings-1.yml
chassing: Good idea. I left a note in [4b8ae33](https://github.com/app-sre/qontract-schemas/pull/562/commits/4b8ae3394a8575b8ca5013a4ee0997412447dfac)

PR #574, opened 2024-01-18
APPSRE-8770 allow DT canada region

fishi0x01:


fishi0x01:
@OliviaHY do you think we could remove this enum? This would need adjustment every time we onboard new region and Im not sure the gained type stability here is worth it (as of now we do not know the region names yet :( )

PR #576, opened 2024-01-23
APPSRE-8660 adjust schema for more self-service

fishi0x01:
We need to accommodate for 2 token:
- the bootstrapAPIToken is used to generate new API tokens for HS
- the automationAPIToken is used by app-interface automations to interact with dynatrace API

Note, that none of the tokens is required. I.e., a dynatrace environment is not necessarily tied to a HS region and/or app-interface automation.

Further, we remove the region enum, as onboarding a new region requires a change to the schema which is not self-servicable.


maorfr (approved):

fishi0x01:
Changing the scope of this change:

- remove enum for better self-service
- add description to bootstrapToken field

Our current automation does not use dynatrace-environment-1 to get the token - it is setup directly through openshiftResources. Once we enhance our automation, we might also change the token structure here at a later point.

PR #577, opened 2024-01-25
Rename template test to resource template test

janboll:
Template test collides with future addition of general templating and it 's comming addition of template tests

[APPSRE-8710](https://issues.redhat.com/browse/APPSRE-8710)


hemslo (approved):
Note this is a breaking change, requires coordinate promotion with qontract reconcile side.

PR #580, opened 2024-01-29
add support for terraform resource aws-iam-role role policy

maorfr:
part of https://issues.redhat.com/browse/APPSRE-8794

in order to create and attach policies instead of using inline policy (or, in addition).


hemslo (approved):
LGTM!

patjlm:
LGTM, would it make sense to have the policy defined as a resource file?

PR #581, opened 2024-01-29
add support for terraform lifecycle

maorfr:
add schema support for the lifecycle meta-argument: https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle

this PR only adds support for the `aws-iam-role` provider.


hemslo (approved):
LGTM!

PR #582, opened 2024-01-29
Add run_order to integration pr_check

hemslo:
Add `run_order` to `pr_check` section for integration, so some integrations can be grouped to run before others.

[APPSRE-8759](https://issues.redhat.com/browse/APPSRE-8759)

### DIFF
--- a/schemas/app-sre/integration-1.yml
+++ b/schemas/app-sre/integration-1.yml
@@ -58,6 +58,8 @@ properties:
       check_only_affected_shards:
         type: boolean
         description: "defaults to false"
+      run_order:
+        type: integer
     required:
     - cmd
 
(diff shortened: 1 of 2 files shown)


chassing (approved):

janboll (approved):

schemas/app-sre/integration-1.yml
```diff
@@ -58,6 +58,8 @@ properties:
       check_only_affected_shards:
         type: boolean
         description: "defaults to false"
+      run_order:
```
janboll, line 61: Is it possible to restrict range? To match supported implementation.

janboll: I see, current implementation does not restrict the numbers. Guess this is not required then

PR #591, opened 2024-02-15
introduce app-interface-settings custom messages

maorfr:
part of https://issues.redhat.com/browse/APPSRE-9820

with this change we can send custom messages to integrations instead of hard coding them in qontract-reconcile.

### DIFF
--- a/schemas/app-interface/app-interface-settings-1.yml
+++ b/schemas/app-interface/app-interface-settings-1.yml
@@ -210,6 +210,22 @@ properties:
     - defaultIssueType
     - defaultPriority
     - defaultReopenState
+  customMessages:
+    type: array
+    description: custom messages to be used in integrations
+    items:
+      type: object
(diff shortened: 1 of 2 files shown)


schemas/app-interface/app-interface-settings-1.yml
```diff
@@ -210,6 +210,22 @@ properties:
...
+    items:
+      type: object
+      additionalProperties: false
+      properties:
+        id:
+          type: string
+          description: unique id of the message
+        content:
+          type: string
+          description: content of the message
```
patjlm, line 225: Is that a Jinja templated string? I think that would be cool to be able to apply context info in messages.

patjlm: I guess this depends on the id being defined and how it's used in the code of course

patjlm (approved):

PR #592, opened 2024-02-15
Add Templating Collection Schema

janboll:
Add template-collection schema, makes it possible to generate datafiles from a collection of templates by passing in variables.


hemslo (approved):
LGTM!

PR #596, opened 2024-02-28
aws account manager

chassing:
Add AWS account manager-related schemas.

Design doc: [appsre-aws-account-management](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/appsre-aws-account-management.md)


chassing:
@hemslo PTAL here as well 🙇

PR #599, opened 2024-03-05
add backref from namespace to gabi instances

maorfr:
this will enable self-service for instances in a gabi instance file approved by the namespace (app) owners.


chassing (approved):
nice!

PR #606, opened 2024-03-14
APPSRE-9922: Add network claim schema

lisa:
As it stands, this will not validate because the
`  - { name: inUseBy, type: string, isRequired: false }` is wrong; it should pair with the definition to be properly supportive of the polymorphic nature of allowing multiple things to claim a network. On the topic of claiming a network, vpc-1 and app-1 are used as placeholders and do not necessarily represent the final configuration.

fyi @geoberle

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2541,7 +2541,7 @@ confs:
   fields:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
-  - { name: name, type: string, isRequired: true }
+  - { name: name, type: string, isRequired: true, isSearchable: true, isUnique: true }
   - { name: labels, type: json }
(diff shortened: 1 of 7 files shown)


esron:
Hello Lisa. So, for the time being the "network claimer" will be a new schema called `managed-vpc`. This new schema will have a field that points to a `network` schema to retrieve the cidr_block to create a managed VPC. I think is best to ship those two together.

What do you think, can we team up in this?

lisa:
> Hello Lisa. So, for the time being the "network claimer" will be a new schema called `managed-vpc`. This new schema will have a field that points to a `network` schema to retrieve the cidr_block to create a managed VPC. I think is best to ship those two together.
> 
> What do you think, can we team up in this?

Absolutely! I'm going to close this PR since git pollute my branch while trying to rebase, but let's get together for the new PR.

PR #610, opened 2024-03-20
APPSRE-9922: First part of network reservation

lisa:
The AWSVPC part will change and is left here only to leave the pattern.

@esron Let's work in this PR to accomplish https://github.com/app-sre/qontract-schemas/pull/606#issuecomment-2009619431

I've left the bulk of the PR the same and will leave it for the new schema to modify the pattern established here.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3733,6 +3733,7 @@ confs:
   - { name: awsgroups_v1, type: AWSGroup_v1, isList: true, datafileSchema: /aws/group-1.yml }
   - { name: awsaccounts_v1, type: AWSAccount_v1, isList: true, datafileSchema: /aws/account-1.yml }
   - { name: awsvpcs_v1, type: AWSVPC_v1, isList: true, datafileSchema: /aws/vpc-1.yml }
+  - { name: aws_managed_vpcs_v1, type: AWSManagedVPC_v1, isList: true }
   - { name: cloudflare_accounts_v1, type: CloudflareAccount_v1, isList: true, datafileSchema: /cloudflare/account-1.yml }
   - { name: cna_experimental_provisioners_v1, type: CNAExperimentalProvisioner_v1, isList: true, datafileSchema: /cna/experimental-provisioner-1.yml }
   - { name: clusters_v1, type: Cluster_v1, isList: true, datafileSchema: /openshift/cluster-1.yml }
@@ -4463,4 +4464,4 @@ confs:
 
 - name: NetworkInUseBy_v1
   fields:
-  - { name: vpc, type: AWSVPC_v1 }
+  - { name: vpc, type: AWSManagedVPC_v1}
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1502,6 +1502,7 @@ confs:
   - { name: labels, type: json }
   - { name: name, type: string, isRequired: true }
   - { name: description, type: string }
+  - { name: region, type: string, isRequired: true }
   - { name: cidr_block, type: Network_v1, isRequired: true }
   - { name: account, type: AWSAccount_v1, isRequired: true }
   - { name: subnets, type: VPCRequestSubnets }
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1489,7 +1489,7 @@ confs:
   - { name: region, type: string, isRequired: true }
   - { name: subnets, type: AWSSubnet_v1, isList: true }
 
-- name: VPCRequestSubnets
+- name: VPCRequestSubnets_v1
   fields:
   - { name: private, type: string, isList: true }
   - { name: public, type: string, isList: true }
@@ -1505,7 +1505,7 @@ confs:
   - { name: region, type: string, isRequired: true }
   - { name: cidr_block, type: Network_v1, isRequired: true }
   - { name: account, type: AWSAccount_v1, isRequired: true }
-  - { name: subnets, type: VPCRequestSubnets }
+  - { name: subnets, type: VPCRequestSubnets_v1 }
 
 - name: AWSSubnet_v1
   fields:
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1489,20 +1489,22 @@ confs:
   - { name: region, type: string, isRequired: true }
   - { name: subnets, type: AWSSubnet_v1, isList: true }
 
-- name: ManagedVPCCIDRBlock_v1
+- name: VPCRequestSubnets
   fields:
-  - { name: network, type: Network_v1, isRequired: true }
+  - { name: private, type: string, isList: true }
+  - { name: public, type: string, isList: true }
 
-- name: AWSManagedVPC_v1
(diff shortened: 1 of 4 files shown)


esron (approved):
LGTM, even tho I've added some commits to it.

graphql-schemas/schema.yml
```diff
@@ -1430,6 +1431,16 @@ confs:
...
   - { name: subnets, type: AWSSubnet_v1, isList: true }
 
+- name: AWSManagedVPC_v1
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: labels, type: json }
+  - { name: name, type: string, isRequired: true }
+  - { name: description, type: string }
+  - { name: network, type: Network_v1, isRequired: true }
```
OliviaHY, line 1441: Is the name network better than `cidr_block:` in `/aws/vpc-1.yml`? In my opinion Network is more vague and doesn't capture the core of the information. Another thing is that we will be having both vpc schemas for a while, would it better to have some parity between them?

graphql-schemas/schema.yml
```diff
@@ -4378,3 +4390,19 @@ confs:
   fields:
   - { name: name, type: string, isRequired: true }
   - { name: query, type: string, isRequired: true }
+
+- name: Network_v1
```
OliviaHY, line 4394: Is there any field need to be unique? For example when I want to request a network, how do I know if there is a yaml file exist for this network already?

schemas/app-sre/network-1.yml
```diff
@@ -0,0 +1,49 @@
...
+    oneOf:
+    - required:
+      - vpc
+  networkAddress:
+    description: "The network address in CIDR format"
+    type: string
+    # Naive regeg; validate in code before use.
+    pattern: "^\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\/\\d+$"
+  parentNetwork:
+    description: "A reference to this network's parent network. If this is a 'root' network, it will have no parent."
```
OliviaHY, line 38: Nice.

graphql-schemas/schema.yml
```diff
@@ -1430,6 +1431,16 @@ confs:
...
 
+- name: AWSManagedVPC_v1
+  fields:
+  - { name: schema, type: string, isRequired: true }
+  - { name: path, type: string, isRequired: true }
+  - { name: labels, type: json }
+  - { name: name, type: string, isRequired: true }
+  - { name: description, type: string }
+  - { name: network, type: Network_v1, isRequired: true }
+  - { name: account, type: AWSAccount_v1, isRequired: true }
```
OliviaHY, line 1442: We need vpc_id to be optional for integrations that depending on it.

schemas/aws/managed-vpc-1.yml
```diff
@@ -0,0 +1,27 @@
...
+  name:
+    type: string
+  description:
+    type: string
+  network:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$scemaRef": "/app-sre/network-1.yml"
+  account:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef": "/aws/account-1.yml"
```
OliviaHY, line 23: vpc_id field is needed.

graphql-schemas/schema.yml
OliviaHY: subnets field too. terraform_tgw_attachments depends on it

esron: Sorry, haven't we decide to create a new schema that doesn't need vpc_id and fetch them from AWS by using a data source in the integrations that needs them?

OliviaHY: You are right. vpc_id isn't needed. That discussion reminded it for me.

graphql-schemas/schema.yml
lisa: I reckon the network address should be unique. I'll make that change.

graphql-schemas/schema.yml
esron: `network` is a reference to the network schema I prefer this because using `cidr_block` can be confusing for someone that didn't participated in the discussions about that.

A developer in the future can think it is weird to have `cidr_block` being a string in a schema and an object in another. More so for similar schemas.

schemas/aws/managed-vpc-1.yml
```diff
@@ -0,0 +1,27 @@
...
+"$schema": /metaschema-1.json
+version: '1.0'
+type: object
+
+additionalProperties: false
+properties:
+  "$schema":
+    type: string
+    enum:
+    - /aws/managed-vpc-1.yml
```
geoberle, line 11: this schema looks a lot like a "request" object similar to what @chassing is doing with AWS account requests.

i'm not sure about your plan but to me the flow of events to create a VPC could be as follows:
* create a `/aws/managed-vpc-1.yml` 
* an integration picks it up, creates a VPC in the account
* the integration then creates an MR to introduce a datafile of schema `/aws/vpc-1.yml` with all the details like VPC IDs
* we keep using `/aws/vpc-1.yml` in all kinds of places where we need to reference VPCs

was that the plan?

lisa:
I have put a hold on this PR because we will be splitting out the network changes from the changes Esron are making. There will need to be some "git-fu" to make sure the changesets are separated in a way that makes sense.

lisa:
> I have put a hold on this PR because we will be splitting out the network changes from the changes Esron are making. There will need to be some "git-fu" to make sure the changesets are separated in a way that makes sense.

I'll take a different approach with the split. Instead of closing this PR, let's leave it open because it has excellent discussion around Esron's changes. The network reservation stuff has been put into its own PR https://github.com/app-sre/qontract-schemas/pull/613. I don't intend to "pull back" the network stuff from this PR because if #613 is merged first, then the merge conflicts between that PR anad this one will be nil to easily manageable.

schemas/aws/managed-vpc-1.yml
esron: That was not the plan at that point. But now it is. Thanks for the insight. I'm going to rename things to make this more clear.

graphql-schemas/schema.yml
```diff
@@ -4448,4 +4467,4 @@ confs:
 
 - name: NetworkInUseBy_v1
   fields:
-  - { name: placeholder, type: string }
+  - { name: vpc, type: VPCRequest_v1}
```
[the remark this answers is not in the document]
lisa: @OliviaHY The plan is that we will follow the same pattern as elsewhere in the schema where subsequent claimants will be enumerated here as well and the integration(s) will need to check the field(s) it cares about for `nil` before use.

graphql-schemas/schema.yml
```diff
@@ -1484,6 +1484,24 @@ confs:
   - { name: region, type: string, isRequired: true }
   - { name: subnets, type: AWSSubnet_v1, isList: true }
 
+- name: VPCRequestSubnets
+  fields:
+  - { name: private, type: string, isList: true }
```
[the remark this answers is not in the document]
esron: Thanks, I'll add that to my notes.

PR #613, opened 2024-03-26
APPSRE-9922: Add network reservation schema

lisa:
This work is intended to be logically split from
https://github.com/app-sre/qontract-schemas/pull/610 which has these changes as well as changes that Esron are working on for a related card. The reason for the split is that these schema changes are, and can be independent from Esron's work.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1418,6 +1418,7 @@ confs:
   - { name: policy, type: json, isRequired: true }
 
 - name: AWSVPC_v1
+  datafile: /aws/vpc-1.yml
   fields:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -4378,3 +4380,18 @@ confs:
...
+  - { name: path, type: string, isRequired: true }
+  - { name: name, type: string, isRequired: true }
+  - { name: description, type: string, isRequired: false }
+  - { name: inUseBy, type: NetworkInUseBy_v1 }
+  - { name: networkAddress, type: string, isRequired: true, isUnique: true }
+  - { name: region, type: string, isRequired: false }
+  - { name: parentNetwork, type: Network_v1, isRequired: false }
+
+- name: NetworkInUseBy_v1
+  fields:
```
esron, line 4397: I think we are missing the fields here.

lisa: > I think we are missing the fields here.

Are we? The other PR will add it. I don't want to add it here because I'm not adding it to the schema. Nothing uses this so it shouldn't be a problem?

esron: Yeah, but not adding anything under fields is failing PR check.

lisa: Hmm, I'm not sure what to add for it.

lisa: I added a placeholder field to hopefully make CI play nice. It will cause a merge conflict for #610, but hopefully it's not very painful to resolve.

esron: Good.

esron (approved):

PR #615, opened 2024-04-03
Add a Flag to enable managing resources with ERv2

lechuk47:
Add a flag to manage resources with ERv2. 
* Roles and RDS for now.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2058,6 +2058,7 @@ confs:
   - { name: annotations, type: json }
   - { name: event_notifications, type: AWSRDSEventNotification_v1, isList: true}
   - { name: data_classification, type: AWSRDSDataClassification_v1 }
+  - { name: managed_by_erv2, type: boolean}
 
 - name: NamespaceTerraformResourceS3_v1
   interface: NamespaceTerraformResourceAWS_v1
@@ -2125,6 +2126,7 @@ confs:
   - { name: output_format, type: NamespaceTerraformResourceOutputFormat_v1, isInterface: true }
   - { name: annotations, type: json }
   - { name: lifecycle, type: NamespaceTerraformResourceLifecycle_v1 }
+  - { name: managed_by_erv2, type: boolean}
 
 - name: NamespaceTerraformResourceElastiCache_v1
   interface: NamespaceTerraformResourceAWS_v1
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -2058,6 +2058,7 @@ confs:
   - { name: annotations, type: json }
   - { name: event_notifications, type: AWSRDSEventNotification_v1, isList: true}
   - { name: data_classification, type: AWSRDSDataClassification_v1 }
+  - { name: managed_by_erv2, type: boolean}
```
hemslo, line 2061: do you think it worth using `provider_version: v1` or `provider_version: v2` to manage this without mention implementation like `erv2`?

And we need json schema changes too.

lechuk47: I see this as a temporary solution/experiment but I also like your idea. I'm not sure if `provider_version` is the more appropriate name though.

lechuk47: I added the json schemas. I think for now is better to stick to the flag approach because it's less invasive and way clearer. 
If the migration needs the `provider_version` we can move to it afterward. Is that ok?

rporres: I thought it was going to be a different type of external resource, the same way that we have multiple schemas for different external resources we can have a new one for a terraform resource managed by a different mechanism... Maybe it is overengineered, though...

If not, I'm ok with the flag approach.

hemslo: if this just a temp flag, then it's ok, I thought we want make the whole config versioned.

hemslo (approved):
LGTM!

graphql-schemas/schema.yml
rporres: yes, I agree

rporres (approved):
LGTM with this being a temp flag

graphql-schemas/schema.yml
lechuk47: To be clear, the idea is to migrate resources to ErV2 once a module is battle-tested. Handling the same "provider" with both mechanisms was not the intention. 

The pattern would be: 
* Test the module setting the erv2 flag in some resources. 
* Once we are confident in the module, Migrate resources to ErV2 and exclude the "provider" in the ErV1 query. 

This would end with the decommission of terraform-resources. I will put this in the milestone doc, we can discuss this further there if you wish.

PR #617, opened 2024-04-04
Fix change VPC request name to identifier

esron:
Name is good in app-interface universe but for a Terraform resource identifier has better restrictions. The same identifier can serve as a `Name` tag.


lisa:
/lgtm

seems reasonable to me. :shipit:

jfchevrette (approved):

PR #618, opened 2024-04-05
Add external resources required schemas

lechuk47:
Adds required schemas for the external resources v2 integration


hemslo (approved):
LGTM!

PR #619, opened 2024-04-08
add escalation policy reference to acs policy

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10043

we'll leverage the escalation policy to automate notifications via ACS

PR #620, opened 2024-04-08
expose ACS defaultFieldsJson via escalation policy jiraCustomFieldsJSON

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10043

following up on adding an escalation policy reference to ACS policies in #619

ACS docs: https://docs.openshift.com/acs/4.4/integration/integrate-with-jira.html#jira-configuring-acs_integrate-with-jira

with this PR we expose the ACS option in a raw manner to allow initial flexibility. we'll see usage patterns and perhaps abstract it in the future.

### DIFF
--- a/schemas/app-sre/escalation-policy-1.yml
+++ b/schemas/app-sre/escalation-policy-1.yml
@@ -45,6 +45,9 @@ properties:
       jiraComponent:
         type: string
         description: component to set on jira tickets
+      jiraCustomFieldsJSON:
+        type: object
+        description: JSON values for custom fields in jira
       nextEscalationPolicy:
         "$ref": "/common-1.json#/definitions/crossref"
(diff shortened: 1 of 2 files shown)


schemas/app-sre/escalation-policy-1.yml
```diff
@@ -45,6 +45,9 @@ properties:
       jiraComponent:
         type: string
         description: component to set on jira tickets
+      jiraCustomFieldsJSON:
```
rporres, line 48: If this is only used in ACS tickets context, it looks weird that it is something that apparently can affect any ticket. If we feel that this may evolve in the future, then I think it makes more sense for it to have a name specific to ACS.

maorfr:
closing in favor of additional implementation in https://github.com/app-sre/qontract-reconcile/pull/4246

PR #625, opened 2024-04-11
acs policy introduce integrations

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10043

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/102231

As `notifiers` is a single form of an integration, we will add a new field to the schema: `integrations`. This object will include a new `notifiers` section, which will have a field for each notifier integration we want to implement, starting with `jira`.

The `integrations.notifiers.jira` field will include `escalationPolicy`, a reference to an Escalation Policy, which will be used to collect the required information to manage Notifiers. Specifically, the first Jira board referenced under `channels` will be used to define the default project, Severity to Priority mappings, and more.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -688,11 +688,23 @@ confs:
   - { name: name, type: string, isRequired: true }
   - { name: description, type: string }
   - { name: notifiers, type: string, isList: true }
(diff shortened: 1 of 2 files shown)


chassing (approved):
LGTM

PR #627, opened 2024-04-12
Globally prefer podman and better cross platform support

lisa:
There were some targets in the Makefile that were hardcoded to use the docker binary. They have been changed to use `$(CONTAINER_ENGINE)`, which will prefer podman and fall back to docker.

In order to better support non-Linux platforms, some auto-detection is done for the OS and if Linux is detected, container volumes will be mounted with `:z` to facilicate SELinux labels (which are not supported on MacOS with Podman). There are two ways to override the auto-detection. To pretend to _not_ be Linux: `make UNAME=Unknown` (or whichever platform). To remain Linux (for example) but disable SELinux mount: `make SELINUX_MOUNT_CHAR=`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,12 @@ BUNDLE_FILENAME ?= data.json
 PWD := $(shell pwd)
 GIT_COMMIT := $(shell git rev-parse HEAD)
 GIT_COMMIT_TIMESTAMP := $(shell git log -1 --format=%ct $(GIT_COMMIT))
+UNAME ?= $(shell uname -s || echo Unknown)
+ifeq ($(UNAME), Linux)
+SELINUX_MOUNT_CHAR = :z
+else
+SELINUX_MOUNT_CHAR =
+endif
 
 ifneq (,$(wildcard $(CURDIR)/.docker))
 	DOCKER_CONF := $(CURDIR)/.docker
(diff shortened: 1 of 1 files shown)


cubismod (approved):
LGTM!

Makefile
```diff
@@ -25,33 +31,33 @@ help: ## Prints help for targets with comments
...
 	mkdir -p $(OUTPUT_DIR) fake_data fake_resources
 	@$(CONTAINER_ENGINE) run --rm \
-		-v $(PWD)/schemas:/schemas:z \
-		-v $(PWD)/graphql-schemas:/graphql:z \
-		-v $(PWD)/fake_data:/data:z \
-		-v $(PWD)/fake_resources:/resources:z \
+		-v $(PWD)/schemas:/schemas$(SELINUX_MOUNT_CHAR) \
+		-v $(PWD)/graphql-schemas:/graphql$(SELINUX_MOUNT_CHAR) \
+		-v $(PWD)/fake_data:/data$(SELINUX_MOUNT_CHAR) \
+		-v $(PWD)/fake_resources:/resources$(SELINUX_MOUNT_CHAR) \
```
hemslo, line 47: `:z` means to share with multiple containers, I don't see where are we using that feature, all volume binds in this file are just read only mount, a simple `:ro` should be good enough.

lisa: @hemslo With regard to `:z`, while you are correct, the container engine facilitates the sharing through SELinux labels, which aren't available outside a SELinux-enabled environment (such as MacOS).

If we want to change the flag to `:ro` later, we can do that at a later date. I want this change to be minimal in scope without potentially breaking changes.

hemslo: Then can we only keep `@$(CONTAINER_ENGINE)` changes in this pr, or just remove `:z` in both cases? `:ro` is a good to have suffix to indicate readonly, then that system check is also not needed anymore, seems small enough to be included all in one change.

PR #630, opened 2024-04-16
♻️ Move availability zones to its own list

esron:
To be able to use azs the same way we use in the [Terraform VPC Module](https://registry.terraform.io/modules/terraform-aws-modules/vpc/aws/latest) I've moved them to their own list.


hemslo (approved):
LGTM!

PR #639, opened 2024-04-29
:recycle: /access/role-1.yml: refactor ldapGroup attribute

chassing:
The goal is to enable our LDAP groups to be authZ groups for Bitwarden collections. Rover group requirements are:

* Rover group notes must contain the string `@BW@` somewhere
* At least 2 owners are set

This PR adds `notes` and `membersAreOwners` (both optional) to the `role.ldapGroup` attribute. `notes` will be Rover group notes, and `membersAreOwners: true` will also grant the `owners` permission to all group members.

Tickets
* [APPSRE-10201](https://issues.redhat.com/browse/APPSRE-10201)
* [APPSRE-8399](https://issues.redhat.com/browse/APPSRE-8399)

### DIFF
--- a/schemas/access/role-1.yml
+++ b/schemas/access/role-1.yml
@@ -177,13 +177,13 @@ properties:
     properties:
       name:
         "$ref": "/common-1.json#/definitions/ldapGroupName"
-        description: The name of the LDAP group
+        description: The name of the LDAP/Rover group
       notes:
         type: string
-        description: Notes added to the LDAP group
+        description: Notes added to the LDAP/Rover group
       membersAreOwners:
         type: boolean
-        description: If true, all members of the LDAP group will be owners as well
+        description: Grant "owner" permission to all members of the LDAP/Rover group
     required:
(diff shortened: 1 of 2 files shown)


schemas/access/role-1.yml
```diff
@@ -173,7 +173,19 @@ properties:
...
   ldapGroup:
-    "$ref": "/common-1.json#/definitions/ldapGroupName"
+    type: object
+    properties:
+      name:
+        "$ref": "/common-1.json#/definitions/ldapGroupName"
+        description: The name of the LDAP group
+      notes:
+        type: string
+        description: Notes added to the LDAP group
```
fishi0x01, line 183: nit: Should we add here that this is targeted for Rover groups?

fishi0x01 (approved):

schemas/access/role-1.yml
chassing: Mentioned rover in the last commit too

PR #642, opened 2024-05-01
Add field to code component to specify image build job URL

bsquizz:
Since we have image builds occurring on different systems at this point (not always on Jenkins), it would be ideal to indicate on a code component where you can find its build job. We plan to reference this field on the developer portal.

### DIFF
--- a/schemas/app-sre/app-1.yml
+++ b/schemas/app-sre/app-1.yml
@@ -295,20 +295,20 @@ properties:
             sourceProject:
               type: object
               properties:
-                name: 
+                name:
                   type: string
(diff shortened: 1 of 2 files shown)


schemas/app-sre/app-1.yml
```diff
@@ -295,20 +295,20 @@ properties:
             sourceProject:
               type: object
               properties:
-                name:
```
bsquizz, line 298: Looks like my IDE just removed some trailing whitespace here...

maorfr (approved):

PR #647, opened 2024-05-06
statuspage-maintenances add more fields/features

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10224

adding a page reference to a maintenance for statuspage provider. also adding self-service for the schema.

design doc: https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/maintenances.md

### DIFF
--- a/schemas/app-sre/maintenance-1.yml
+++ b/schemas/app-sre/maintenance-1.yml
@@ -44,6 +44,9 @@ properties:
           type: string
           enum:
           - statuspage
+        page:
+          "$ref": "/common-1.json#/definitions/crossref"
+          "$schemaRef": "/dependencies/status-page-1.yml"
         announceAt:
           type: string
           format: date-time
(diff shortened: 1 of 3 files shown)


schemas/app-sre/maintenance-1.yml
```diff
@@ -82,6 +88,7 @@ properties:
           type: boolean
       required:
       - provider
+      - page
```
patjlm, line 91: For my understanding, this is a new required field that will be used to actually notify subscribers (list them or use status page API of some sort), right?

it sounds weird to have it required with the provider pattern, but as we have only maintenance announcements of type statuspage for now, it's fine of course.

patjlm (approved):

schemas/app-sre/maintenance-1.yml
maorfr: this is exactly the provider pattern. we are under `oneOf`: https://github.com/app-sre/qontract-schemas/pull/647/files#diff-8954fe83fbfc14d023e6ab1d5e0eca9b03c919e4d0dfbcb2d2eec677360c0a3bR67

PR #653, opened 2024-05-20
enable self-service for slack-workspace

maorfr:


patjlm (approved):

maorfr:
no, it's about the ability to even self-service things within an existing workspace

PR #663, opened 2024-06-05
:sparkles: aws-account: manage the security contact

chassing:
Introduce the optional `securityContact` field to `/aws/account-1.yml`. 

Ticket: [APPSRE-10353](https://issues.redhat.com/browse/APPSRE-10353)

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1281,6 +1281,7 @@ confs:
   - { name: qontractReconcileVersion, type: string }
   - { name: terraformUsername, type: string }
   - { name: accountOwners, type: Owner_v1, isList: true, isRequired: true }
+  - { name: securityContact, type: AwsContact_v1 }
   - { name: automationToken, type: VaultSecret_v1, isRequired: true }
   - { name: automationRole, type: AWSAutomationRole_v1 }
   - { name: garbageCollection, type: boolean }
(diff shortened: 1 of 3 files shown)


graphql-schemas/schema.yml
```diff
@@ -1281,6 +1281,7 @@ confs:
   - { name: qontractReconcileVersion, type: string }
   - { name: terraformUsername, type: string }
   - { name: accountOwners, type: Owner_v1, isList: true, isRequired: true }
+  - { name: securityContact, type: AwsContact_v1 }
```
esron, line 1284: Nit: s/Aws/AWS/g

chassing: Fixed in last commit

schemas/aws/account-1.yml
```diff
@@ -43,6 +43,24 @@ properties:
...
+        type: string
+        description: "The name of the security contact."
+      title:
+        type: string
+        description: "The title of the security contact. Default use the name."
+      email:
+        type: string
+        format: email
+      phoneNumber:
+        type: string
```
rporres, line 59: should we enforce something here, like `^+\d+` with a max length of 16? Just to make sure that we use proper international numbers here...

rporres: or at least what AWS accepts: https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-update-contact-alternate.html#manage-acct-update-contact-alternate-requirements

chassing: I added a very basic and simple phone number check. Should be enough for our use-case.

esron (approved):
LGTM

PR #667, opened 2024-06-06
template-renderer: foreach.items loops in template-collections

patjlm:
https://issues.redhat.com/browse/APPSRE-10379

example:
```yaml
forEach:
  items:
  - cluster_name: app-sre-prod-01
    value: value1
  - cluster_name: app-sre-prod-04
    value: value2

variables:
  static:
    foo: FOO
    bar: "This is for value {{ foreach_item.value }}"
  dynamic:
  - name: cluster
    query: |
      {
      clusters: clusters_v1(filter: {
          name:"{{ foreach_item.cluster_name }}"
        })
        {
          name
          consoleUrl
        }
      }
```

### DIFF
--- a/schemas/app-interface/template-collection-1.yml
+++ b/schemas/app-interface/template-collection-1.yml
@@ -22,6 +22,14 @@ properties:
     type: string
   enableAutoApproval:
     type: boolean
+  forEach:
+    type: object
+    properties:
+      items:
+        type: array
+        description: object items on which to run this collection
+        items:
+          type: object
   variables:
     type: object
     description: variables to be used in the template
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -4630,9 +4630,14 @@ confs:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
   - { name: enableAutoApproval, type: boolean, isRequired: false }
+  - { name: forEach, type: TemplateCollectionForEach_v1 }
   - { name: description, type: string, isRequired: true }
   - { name: variables, type: TemplateCollectionVariables_v1 }
   - { name: templates, type: Template_v1, isList: true, isRequired: true }
+
+- name: TemplateCollectionForEach_v1
+  fields:
+  - { name: items, type: json, isList: true, isRequired: true }
 
 - name: TemplateCollectionVariables_v1
   fields:


schemas/app-interface/template-collection-1.yml
```diff
@@ -22,6 +22,14 @@ properties:
...
   enableAutoApproval:
     type: boolean
+  forEach:
+    type: object
+    properties:
+      items:
+        type: array
+        description: object items on which to run this collection
+        items:
+          type: object
```
hemslo, line 32: Any other properties under `forEach`?

How about directly make `forEach` as list of object, like 

```yaml
forEach:
- cluster_name: app-sre-prod-01
  value: value1
- cluster_name: app-sre-prod-04
  value: value2
```

hemslo (commented):
How about simplify variable name from `foreach_item` to just `each`? Borrow from terraform [for_each](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each)

```yaml
forEach:
- cluster_name: app-sre-prod-01
  value: value1
- cluster_name: app-sre-prod-04
  value: value2

variables:
  static:
    foo: FOO
    bar: "This is for value {{ each.value }}"
  dynamic:
  - name: cluster
    query: |
      {
      clusters: clusters_v1(filter: {
          name:"{{ each.cluster_name }}"
        })
        {
          name
          consoleUrl
        }
      }
```

And `each` is only known in template collection level, template still only have access to `static` and `dynamic`.

schemas/app-interface/template-collection-1.yml
patjlm: next step would be to add an alternative to `items` based on `query` + `jsonpath`

patjlm:
> How about simplify variable name from `foreach_item` to just `each`? Borrow from terraform [for_each](https://developer.hashicorp.com/terraform/language/meta-arguments/for_each)

I am not absolutely against it, but it looks less obvious to me. Happy to get others' thoughts

> And `each` is only known in template collection level, template still only have access to `static` and `dynamic`.

Why would it not be available to templates? I think that can be handy. One could 
```yaml
forEach:
- var: value1
- var: value2
variables:
  static:
    var: "{{ each.var }}"
```
but for more complex structures, i can become complex

schemas/app-interface/template-collection-1.yml
hemslo: oh, I see, another layer of static and dynamic, right?

hemslo:
> Why would it not be available to templates? I think that can be handy. One could

```yaml
forEach:
- var: value1
- var: value2
variables:
  static:
    var: "{{ each.var }}"
```

what I mean is `forEach` only be used in `variables` resolving for `template-collection` level, when passing down to `template`, `template` doesn't need to know the concept of `forEach`, only need to know `static` and `dynamic` as before, no changes at all. All examples here and in tests are just for `static` and `dynamic` resolving, any other use cases?

Basically no passing [here](https://github.com/app-sre/qontract-reconcile/pull/4429/files#diff-d722c4a425fe398846ecdbccc881ac57ee4291d358c22b4d11b8fae7adc4a8c8R269)

patjlm:
> Basically no passing [here](https://github.com/app-sre/qontract-reconcile/pull/4429/files#diff-d722c4a425fe398846ecdbccc881ac57ee4291d358c22b4d11b8fae7adc4a8c8R269)

Why would i be a problem to pass it down to the template?

hemslo:
> > Basically no passing [here](https://github.com/app-sre/qontract-reconcile/pull/4429/files#diff-d722c4a425fe398846ecdbccc881ac57ee4291d358c22b4d11b8fae7adc4a8c8R269)
> 
> Why would i be a problem to pass it down to the template?

To make `template` simple, currently given a template collection like

```yaml
variables:
  static:
    foo: FOO
    bar: "This is for value {{ each.value }}"
  dynamic:
  - name: cluster
    query: |
      {
      clusters: clusters_v1(filter: {
          name:"{{ each.cluster_name }}"
        })
        {
          name
          consoleUrl
        }
      }
```

I know `{{ static.foo }}` and `{{ dynamic.cluster }}` are available as both `static` and `dynamic` are under `variables`, if we add more keys to it without actual use case, then it may confuse people what variables are available, need additional document, and requires more test cases or template tests to cover complex usage planned [here](https://github.com/app-sre/qontract-schemas/pull/667#discussion_r1631094723). Model it as strict type, template declares variables as an object with `static` and `dynamic` field, template collection provides them. So all the changes for `foreach` can be an enhancement for template collection, keep template untouched.

schemas/app-interface/template-collection-1.yml
patjlm: it would be either one or the other
```yaml
forEach:
  items: [...]
```
or something like
```yaml
forEach:
  query: |...  # gets data from graphql
  jsonpath: ...  # exctracts a list of items from the graphql result
```

patjlm:
@hemslo I renamed the variable to `each` and removed the passing to the individual templates.

I think we may hit some difficulties with more complex object structures in the loop items, but we can always revise that later.

chassing (approved):

PR #673, opened 2024-06-18
make slack permission channels mandatory

maorfr:
it would make sense for every usergroup to define in what channels they should be contacted. specifying channels should be at least mandatory to begin with.


chassing (approved):

chassing:
Do not forget to notify the fedramp guys about this breaking change.

PR #675, opened 2024-06-18
:sparkles: templates: add templateRenderOptions attribute

chassing:
Make the jinja [whitespace control](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control) behavior configurable for a template by introducing `trim_blocks`, `lstrip_blocks`, and `keep_trailing_newline` as `templateRenderOptions`. The default will still be `False` for all three options.


hemslo (approved):
LGTM!

PR #679, opened 2024-06-24
Update identifier pattern in schemas to allow dots in strings

raks-tt:
This change modifies the identifier pattern to accommodate
 strings with dots (.) such as 'raks.tt' in user-1.yml schema.

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -34,7 +34,7 @@
     },
     "identifier": {
       "type": "string",
-      "pattern": "^[A-Za-z0-9][A-Za-z0-9-_]{0,30}[A-Za-z0-9]$"
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9.-_]{0,30}[A-Za-z0-9]$"
     },
     "identifierLowercase64": {
       "type": "string",


rporres:
Can you add some context to this proposal? What's the problem you're trying to solve?

raks-tt:
> Can you add some context to this proposal? What's the problem you're trying to solve?

@rporres - The issue is we use this in the Konflux repo for validating the schemas. So if we have a string with a "." for eg: raks.tt is my username in quay, the Job will fail. 

Here is the errror:
```
"error": "'raks.tt' does not match '^[A-Za-z0-9][A-Za-z0-9-_]{0,30}[A-Za-z0-9]$'\n\nFailed validating 'pattern' in schema['properties']['quay_username']:\n    {'pattern': '^[A-Za-z0-9][A-Za-z0-9-_]{0,30}[A-Za-z0-9]$',\n     'type': 'string'}\n\nOn instance['quay_username']:\n    'raks.tt'",

            "schema_url": "/access/user-1.yml"
```

rporres:
If I understand correctly, the problem you're trying to solve is the `quay_username` property from the `/access/user-1.yml` schema. Apparently it allows for dots whereas we don't.

The problem with your patch is that it modifies a type definition that's used in many places, so we cannot accept it as it is. You should just change the `quay_username` property. You can define a pattern that a property must comply with directly in the schema. You can see an example of how to do it in the same `schemas/access/user-1.yml` file, e.g. the `aws_username`.

Do you have any documentation of the accepted characters of the usernames in quay?

PR #681, opened 2024-06-26
add aws billing account to aws account

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10457

to be able to use a specified aws account as the billing account in a ROSA HCP cluster provisioning flow


geoberle (approved):

mehfuz:
Nice...i had a question about what is used if this is not provided?...[this](https://github.com/app-sre/qontract-reconcile/blob/3608f6e709cec5a48ebd784e3fed2c8bfb5e6e11/tools/qontract_cli.py#L1363) answered that question..

PR #682, opened 2024-06-27
Add a new field subnets for machine pool 

OliviaHY:
They can span across more than one subnet for multi-az cluster. This way it's backwards compatible until we are ready to migrate to this new field.


OliviaHY:
So this will not be useful as we remove the default machine pool which creates nodes in all subnets then create one machine pool designated for each AZ/subnet, and there is no way to provide more than one subnet ids when creating a machine pool through ROSA CLI and OCM API.

PR #685, opened 2024-07-08
Allow pulling image from Dynatrace registry.

OliviaHY:
To support ActiveGate images coming from each Dynatrace tenant, we'd like to allow pulling directly from there. 

Context: [APPSRE-10517](https://issues.redhat.com/browse/APPSRE-10517)


maorfr (approved):
LGTM for the idea to pull directly from dynatrace.

the regex has a problem though, please fix it before merging.
for example, it doesn't match `ddl70254.live.dynatrace.com/linux/activegate`

OliviaHY:
> LGTM for the idea to pull directly from dynatrace.
> 
> the regex has a problem though, please fix it before merging. for example, it doesn't match `ddl70254.live.dynatrace.com/linux/activegate`

Thank you. I made a mistake while tweaking it, it's correct now: 
image

PR #691, opened 2024-07-16
Add an option to tf-repo enabling a user to force a re-run of the executor

cubismod:
[APPSRE-8705](https://issues.redhat.com/browse/APPSRE-8705)

Part of the work around refining Terraform Repo for usage in commercial. This adds an option to the schema to force a re-run of tf-repo which has been a longstanding oversight. Currently, to force a re-run of the executor, a user needs to make a commit to the upstream Git repository holding the Terraform source files that executor runs. From there that user has to update the SHA in the schema file. This MR adds an easier option for forcing a re-run by allowing a user to submit an MR to the schema file with a timestamp which will then trigger an executor run. 

Due to how the Tekton pipeline is architected, re-running the pipeline itself with Tekton will be a no-op operation since tf-repo's state is updated so that each update to the schema file corresponds with one execution in Tekton, preventing a case where a pipeline is stuck constantly re-running forever and failing.


jonmosco (approved):
LGTM

PR #692, opened 2024-07-25
saas file expose validate planned data boolean

maorfr:
to be able to define it as `False` and simplify initial bootstrap of resources.


mehfuz (approved):
lgtm

PR #698, opened 2024-08-05
Draft: Pyrra specific SLO schema addition

suzana-nesic:


suzana-nesic:
Reworking this as it's way too granular and should be a bit more generic

PR #700, opened 2024-08-13
APPSRE-7516 flag for saas files holding tests

fishi0x01:
Cleaning up tech debt. We currently use `publishJobLogs` flag in saas files to identify whether a saas file holds test jobs. This also has effects on whether target_config_hash is used. However, this is very hidden knowledge and caused confusion in the past.

This PR introduces the flag `hasTestJobs` for saas files, which explicitly states whether tests are part of this saas file. This flag will then be used by integrations (e.g., saasherder, SAPM).


maorfr (approved):
i'd call the field `testing`, as if to indicate this is a testing saas file.
can't help myself, sorry. please feel free to ignore the proposal.
do add a description :pray:

PR #701, opened 2024-08-13
preparation to ping author on MR

fishi0x01:
Porting what we do in qr.


maorfr (approved):
LGTM

we need the qr implementation first

fishi0x01:
Corresponding QR change is here https://github.com/app-sre/qontract-reconcile/pull/4618

PR #704, opened 2024-08-16
Revert "APPSRE-7516 flag for saas files holding tests (#700)"

fishi0x01:
This reverts commit 5db9e797fa44c122ebd8910d8a9e237a66ee7704.

After discussion with @hemslo we decided to refine this a little. This feature is not used yet so it is safe to remove from schema.

I will open a follow up PR


fishi0x01:
Follow up on https://github.com/app-sre/qontract-schemas/pull/700

PR #705, opened 2024-08-16
APPSRE-7516 redeploy on config change

fishi0x01:
Explicit flag to declare whether a target should be re-deployed if there was a config change in the publisher. Previously, this was defined saas file wide via `publishJobLogs` flags.

Follow-up on https://github.com/app-sre/qontract-schemas/pull/700

cc @hemslo 

If desired, we can also have this flag on saas file level instead (i.e., enable/disable for all targets in a saas file).

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3040,6 +3040,7 @@ confs:
 - name: SaasResourceTemplateTargetPromotion_v1
   fields:
   - { name: auto, type: boolean }
+  - { name: reDeployOnPublisherConfigChange, type: boolean }
   - { name: publish, type: string, isList: true }
   - { name: subscribe, type: string, isList: true }
   - { name: soakDays, type: int }
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -3040,7 +3040,7 @@ confs:
 - name: SaasResourceTemplateTargetPromotion_v1
   fields:
   - { name: auto, type: boolean }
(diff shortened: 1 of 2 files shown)


hemslo (approved):
LGTM!

rporres (approved):

graphql-schemas/schema.yml
```diff
@@ -3040,6 +3040,7 @@ confs:
 - name: SaasResourceTemplateTargetPromotion_v1
   fields:
   - { name: auto, type: boolean }
+  - { name: reDeployOnPublisherConfigChange, type: boolean }
```
rporres, line 3043: apologies for the late review, I've just seen this...

I would go for `redeploy` instead of `reDeploy` since redeploy is actually a word: https://www.merriam-webster.com/dictionary/redeploy

```suggestion
  - { name: redeployOnPublisherConfigChange, type: boolean }
```

schemas/app-sre/saas-file-target-1.yml
```diff
@@ -44,6 +44,11 @@ properties:
     properties:
       auto:
         type: boolean
+      reDeployOnPublisherConfigChange:
```
rporres, line 47:
```suggestion
      redeployOnPublisherConfigChange:
```

fishi0x01: Fixed everywhere

graphql-schemas/schema.yml
fishi0x01: Fixed

rporres (approved):

PR #706, opened 2024-08-16
APPSRE-10767 displayName for saas file

fishi0x01:
Due to saas progressive delivery efforts, it is likely that saas file names are more and more used in visualizations. That probably leads to more ppl trying to rename saas files. However, renaming saas files is hard and essentially means migrating targets from one saas to another (including auto promotions being stuck in current ref ...).

We introduce extra field display_name that can be used by visualizations (like backstage and our own grafana dashboard). I.e., we can keep `name` stable as currently we use it like an id.

Note, that `displayName` is required, so any external tooling can be sure it is available. We can add a simple migration script in app-interface to add this to all saas files.


rporres (approved):
nice!

PR #709, opened 2024-08-23
APPSRE-10782 add optional TGW conection cidrBlocks field

patjlm:
https://issues.redhat.com/browse/APPSRE-10782

This will allow to route multiple cidr blocks towards a TGW

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1025,6 +1025,7 @@ confs:
   - { name: manageRoute53Associations, type: boolean }
   - { name: allowPrivateHcpApiAccess, type: boolean }
   - { name: cidrBlock, type: string }
+  - { name: cidrBlocks, type: string, isList: true }
   - { name: delete, type: boolean }
   - { name: assumeRole, type: string }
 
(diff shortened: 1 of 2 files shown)


rporres:
What's the plan here? move all cidrBlock to cidrBlocks and remove cidrBlock in the near future?

patjlm:
> What's the plan here? move all cidrBlock to cidrBlocks and remove cidrBlock in the near future?

that would be best but nor fully required. This needs to be synced with all instances where we use TGW attachements (incl fedramp i suppose).

rporres:
Reading the code I don't think it is feasible to remove `cidrBlock` as I understand that you need to define the `moved` block.

I would then add documentation in the schema to state the usage of `cidrBlock` is deprecated and that `cidrBlocks` should be used instead for new attachments.

patjlm:
Once the resource has been renamed, the `moved` block is not needed and can be removed. At that point we can remove safely `cidrBlock` and move its value within `cidrBlocks`

rporres:
Great, then at some point we can remove cidrBlock from the schema and just use cidrBlocks

As for the deprecation message, I'd just add a note in the `description` field.

rporres (approved):

PR #711, opened 2024-09-08
aws account resetPassword depends on sso being false

maorfr:
part of https://issues.redhat.com/browse/APPSRE-10862

this PR will forbid reset password unless `sso` is set to false.


lechuk47:
I didn't know about this schema feature. 
Nice approach

lechuk47 (approved):

PR #718, opened 2024-10-03
SLOD doc: set prometheusAccess creds as optional

patjlm:
https://issues.redhat.com/browse/APPSRE-8513

Allows unauthenticated local token-refresher svc access, much like what grafana does.

Something like
```yaml
$schema: /app-sre/slo-document-1.yml

name: rosa-classic-fleet-wide

namespaces:
- namespace:
    $ref: /services/rosa-fleet-wide/classic/namespaces/rosa-fleet-wide.yml
  prometheusAccess:
    url: http://osd-token-refresher-prod.app-sre-observability-production.svc.cluster.local
  SLONamespace:
    $ref: /services/rhobs/observatorium-mst/namespaces/telemeter-prod-01/observatorium-mst-production.yml
```

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -4288,8 +4288,8 @@ confs:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
   - { name: url, type: string, isRequired: true }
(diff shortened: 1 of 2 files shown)


rporres:
it should be noted that this has a potential to break easily if cluster is not appsrep09ue1 :(

PR #726, opened 2024-10-29
add app and environment to shared-resources

maorfr:
improve traceability of changes and open room for validations that namespaces use the correct shared resources.


rporres:
I don't understand this... What does app mean in the context of a shared resource? At the end it is just a sort of library of resources to avoid duplication. What happens if the app of the shared resource is different to the app of the namespace where it is used? Same goes to environment.

PR #730, opened 2024-10-31
[tf-repo] Allow for newer versions of TF binary

cubismod:
Allow users to specify newer versions of Terraform in their tf-repo spec file.

[APPSRE-11103](https://issues.redhat.com/browse/APPSRE-11103)


esron (approved):
LGTM

PR #749, opened 2024-12-09
Fix reference for ssl_policy

suzana-nesic:


esron (approved):
LGTM, I've tested locally and it pass on that MR on `app-interface`.

PR #755, opened 2025-01-14
[APPSRE-11112] add must_pass tests to gitlabHousekeeping config

BumbleFeng:


schemas/app-sre/app-1.yml
```diff
@@ -358,6 +358,14 @@ properties:
...
                 required:
                 - role
+            must_pass:
+              type: array
+              items:
+                type: string
+                description: List of tests must pass before merging.
+            persistent_pass:
+              type: boolean
+              description: Persistent pass results for ondemand tests in case of auto rebase.
```
rporres, line 368: This is not in the design-doc... it would be nice to amend it to include it

https://gitlab.cee.redhat.com/service/app-interface/-/blob/068d847cdc82ffd313509fe159328e5b7623c295/docs/app-sre/design-docs/gitlab-ondemand-test.md

BumbleFeng: remove it atm.

PR #756, opened 2025-01-15
APPSRE-11390 fleet labeler schema

fishi0x01:
Schema for the Fleet Labeler Integration.

[Design doc](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/design-docs/fleet-labeler.md?ref_type=heads)

[Example Use](https://gitlab.cee.redhat.com/service/sre-capabilities/-/merge_requests/2223)

Note, I would like to add this schema to main ahead of the integration implementation itself, as it would ease development by not needing to stay in sync with schema changes. Further, it helps to separate implementation into incremental PRs instead of single large PR. The schema does not change any existing behavior and is fully dedicated to fleet labeler, so merging ahead has no real downsides.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1169,6 +1169,25 @@ confs:
   - { name: upgradePolicy, type: ClusterUpgradePolicy_v1 }
   - { name: upgradePolicyTemplate, type: OpenShiftClusterManagerUpgradePolicyTemplate_v1 }
 
+- name: FleetLabelTemplate_v1
+  fields:
+  - { name: path, type: string, isRequired: true, isResource: true, resolveResource: true }
+  - { name: type, type: string }
+  - { name: variables, type: json }
+
+- name: FleetLabelDefault_v1
+  fields:
+  - { name: name, type: string, isRequired: true, isUnique: true }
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -1169,6 +1169,25 @@ confs:
...
+  fields:
+  - { name: path, type: string, isRequired: true, isResource: true, resolveResource: true }
+  - { name: type, type: string }
+  - { name: variables, type: json }
+
+- name: FleetLabelDefault_v1
+  fields:
+  - { name: name, type: string, isRequired: true, isUnique: true }
+  - { name: matchSubscriptionLabels, type: json, isRequired: true }
+  - { name: subscriptionLabelTemplate, type: FleetSubscriptionLabelTemplate_v1 }
```
chassing, line 1182: This attribute is required in the schema but not here. Please also add `isRequired: true`.

chassing (approved):
Just a minor nit.

graphql-schemas/schema.yml
fishi0x01: Oh great catch! Fixed :+1:

PR #760, opened 2025-01-23
elasticache: add service updates related fields

chassing:
Add some service updates related fields to the Elasticache resource files.
Additionally, `auto_minor_version_upgrade` isn't required anymore and will default to `true`.

Ticket: [APPSRE-11432](https://issues.redhat.com/browse/APPSRE-11432)

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -2852,7 +2852,7 @@ confs:
   fields:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
-  - { name: labels, type: json }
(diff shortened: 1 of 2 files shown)


hemslo (approved):
LGTM!

rporres:
what about `automatic_failover_enabled`? Shouldn't it disappear as `auto_minor_version_upgrade` and be always true?

chassing:
> what about `automatic_failover_enabled`? Shouldn't it disappear as `auto_minor_version_upgrade` and be always true?

Good point! I want to tackle the automatic minor version upgrade in an extra ticket. I don't want to confuse and stress everybody about both things simultaneously.

PR #762, opened 2025-01-27
add bound_service_account_namespace_selector to k8s role for vault

jonmosco:
Add `bound_service_account_namespace_selector` to schema.  Required by vault-manager to successfully reconcile k8s auth method roles.

```
[bound_service_account_namespace_selector](https://developer.hashicorp.com/vault/api-docs/auth/kubernetes#bound_service_account_namespace_selector) (string: "") - A label selector for Kubernetes namespaces allowed to access this role. Accepts either a JSON or YAML object. The value should be of type [LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#labelselector-v1-meta). Currently, label selectors with matchExpressions are not supported. To use label selectors, Vault must have permission to read namespaces on the Kubernetes cluster. If set with bound_service_account_namespaces, the conditions are ORed.
```

### DIFF
--- a/schemas/vault-config/role-1.yml
+++ b/schemas/vault-config/role-1.yml
@@ -217,6 +217,8 @@ properties:
           type: array
           items:
             type: string
+        bound_service_account_namespace_selector:
+          type: string
         token_ttl:
           type: string
           pattern: '^\d+$'
@@ -249,6 +251,7 @@ properties:
       - alias_name_source
       - bound_service_account_names
       - bound_service_account_namespaces
+      - bound_service_account_namespace_selector
       - token_ttl
       - token_max_ttl
       - token_explicit_max_ttl
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -346,6 +346,7 @@ confs:
   - { name: audience, type: string }
   - { name: bound_service_account_names, type: string, isList: true, isRequired: true }
   - { name: bound_service_account_namespaces, type: string, isList: true, isRequired: true }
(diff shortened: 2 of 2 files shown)


schemas/vault-config/role-1.yml
```diff
@@ -217,6 +217,8 @@ properties:
           type: array
           items:
             type: string
+        bound_service_account_namespace_selector:
+          type: string
```
patjlm, line 221: shouldn't that be a `LabelSelector` object instead of a string?

patjlm:
Can you reference the corresponding ticket?

jonmosco:
> Can you reference the corresponding ticket?

This work is part of this epic:
https://issues.redhat.com/browse/APPSRE-11373

This was found after using the k8s auth method in FedRamp and vault-manager doing a constant reconciliation.  Investigation of why the reconciliation keeps occurring has pointed to certain attributes missing, this being one of them.  Vault includes the attribute in role definitions by default, even if it is not used, which causes vault-manager to think it needs to perform a reconcile.  

> Is bound_service_account_namespace_selector safe enough. It's pretty easy for anyone to set namespace labels.

This is only for the k8s auth methods role configuration and will be set to an empty string in order to satisfy the reconciliation.  If this proves to be the main reason for vault-manager's behavior, a different approach can be investigated.

jonmosco:
Closing in favor of https://github.com/app-sre/vault-manager/pull/136

This prevents the reconciliation experienced when using k8s auth methods roles.

PR #769, opened 2025-02-05
APPSRE-11390 fleet labeler fixes

fishi0x01:
- we use a single OCM instance. It will cover all current cases. Further, it simplifies logic as we do not need a mapping between clusters and ocms.
- More explicit naming: fleet-label-1 to fleet-label-spec-1. 
- `subscriptionLabels` are a json

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1212,14 +1212,14 @@ confs:
   - { name: provider, type: string, isRequired: true }
   - { name: enforced, type: boolean, isRequired: true }
 
-- name: FleetLabels_v1
-  datafile: /openshift/fleet-labels-1.yml
+- name: FleetLabelsSpec_v1
+  datafile: /openshift/fleet-labels-spec-1.yml
   fields:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
   - { name: name, type: string, isRequired: true, isUnique: true, isSearchable: true }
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -1212,14 +1212,14 @@ confs:
...
-  datafile: /openshift/fleet-labels-1.yml
+- name: FleetLabelsSpec_v1
+  datafile: /openshift/fleet-labels-spec-1.yml
   fields:
   - { name: schema, type: string, isRequired: true }
   - { name: path, type: string, isRequired: true }
   - { name: name, type: string, isRequired: true, isUnique: true, isSearchable: true }
   - { name: description, type: string }
-  - { name: ocms, type: OpenShiftClusterManager_v1, isList: true, isRequired: true }
+  - { name: ocm, type: OpenShiftClusterManager_v1, isRequired: true }
```
hemslo, line 1222: this is a breaking change, will it cause error in integration before schema sync in qontract-reconcile?

fishi0x01: It wont cause error - no integration in prod using this schema

fishi0x01: In general: no change in this PR is used by any integration that is running right now

hemslo: all good then!

hemslo (approved):
LGTM!

PR #772, opened 2025-02-18
APPSRE-11544 remove saas target image

fishi0x01:
Time to say goodbye to `saas_file.resource_template.target.image` in favor of the new `images` (https://github.com/app-sre/qontract-schemas/pull/761).

All streams are aware if this change. App-interface data has been migrated to new `images` already.


hemslo (approved):
LGTM!

PR #773, opened 2025-02-25
APPSRE-11390 store subscription_id in inventory

fishi0x01:
Fleet labeler needs the subscription ID to properly set subscription labels for clusters.


hemslo (approved):
LGTM!

patjlm:
It could also be automatically added, much like the cluster id

PR #774, opened 2025-02-25
add slos field saas file

mehfuz:


gwenneg:
One way to solve that could just consist in adding an optional `cluster` or `namespace` field (I'm not sure which word is most relevant) to https://github.com/app-sre/qontract-schemas/blob/main/schemas/app-sre/slo-document-1.yml to specify where the expression should be evaluated.

mehfuz:
> Hey @mehfuz! I just saw this PR while I was looking for something else in the repo. I'm curious about something: how will the namespace/cluster where the SLO expression is evaluated be determined? For instance, if a team depends on Kafka in AWS (MSK), they might build an SLO on top of metrics that are only available in the `appsre[s|p]11ue1` cluster which is different from the cluster their service is deployed to. Do slo-documents support multi-clusters/namespaces SLOs?

@gwenneg  In case your slo is deployed in a different u can use the `SLONamespace` and `prometheusAccess` so that slo will query on the prometheusAccess.URL and evalute the SLO. Reference

mehfuz:
> One way to solve that could just consist in adding an optional `cluster` or `namespace` field (I'm not sure which word is most relevant) to https://github.com/app-sre/qontract-schemas/blob/main/schemas/app-sre/slo-document-1.yml to specify where the expression should be evaluated.

Refer comment https://github.com/app-sre/qontract-schemas/pull/774#issuecomment-2760787181

graphql-schemas/schema.yml
```diff
@@ -3112,6 +3112,7 @@ confs:
   - { name: secretParameters, type: SaasSecretParameters_v1, isList: true }
   - { name: upstream, type: SaasResourceTemplateTargetUpstream_v1 }
   - { name: images, type: SaasResourceTemplateTargetImage_v1, isList: true }
+  - { name: slos, type: SLODocument_v1 ,isList: true, isRequired: false}
```
chassing, line 3115: The attribute is required here but not in `/app-sre/saas-file-target-1.yml`. Which one is correct?

mehfuz: the attribute is supposed to be optional. Setting `isRequired: false` here should keep it optional right?

chassing: I read `isRequired: true` 🤦

PR #778, opened 2025-03-03
APPSRE-11584 fleet-labeler label synch flag

fishi0x01:
Extra flag for the fleet-labeler integration. 

Disable the synchronization of labels from the subscription to the cluster. This is useful if you want to let fleet labeler render the inventory first and investigate/change any desired labels manually before the first synchronization starts.


rporres:
Since this looks like a dry-run at a spec lebel, does it make sense to call it dryRun to decrease some cognitive load or would it be misleading since we're used to dry runs at an integration level?

fishi0x01:
There are 2 phases for dry-run: inventory synch (MRs to update `clusters` section) and label synch. Would it make sense to rename to `dryRunLabelSynch`?

PR #785, opened 2025-03-19
:sparkles: aws-account: add organizationAccountTags

chassing:
This PR adds `organizationAccountTags` to the `/aws/account-1.yml` schema. This enables the `aws-account-manager` to use this attribute to tag all accounts within a payer account with common tags, e.g. `cost-center: xxx`

Ticket: [APPSRE-11633](https://issues.redhat.com/browse/APPSRE-11633)

### DIFF
--- a/schemas/aws/account-1.yml
+++ b/schemas/aws/account-1.yml
@@ -162,7 +162,7 @@ properties:
       description: "AWS quota limits for this account"
   organizationAccountTags:
     "$ref": "/common-1.json#/definitions/labels"
-    description: "Apply these tags to all accounts in the organization. This is useful for tagging all accounts in the organization with the same tags."
+    description: "For payer accounts only! Apply these tags to all accounts in the organization. This is useful for tagging all accounts in the organization with the same tags."
 dependencies:
   resetPasswords:
     properties:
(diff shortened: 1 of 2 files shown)


schemas/aws/account-1.yml
```diff
@@ -167,6 +170,11 @@ dependencies:
         - false
     required:
     - sso
+  organizationAccountTags:
+    properties:
+      organization:
+        enum:
+        - null
```
hemslo, line 177: why `organizationAccountTags` requires `organization` to be `null`?

chassing: If an account has an `organization` set, it's an organization account; if not, it's a payer account. Only payer accounts can have the `organizationAccountTags` attribute.

hemslo: can we add this limit to `description`? like only available for payer account.

chassing: Sure. Done in [6b2d9c6](https://github.com/app-sre/qontract-schemas/pull/785/commits/6b2d9c6575da13645e46ba6240b78b06320d2158)

hemslo (approved):
LGTM!

PR #786, opened 2025-03-19
aws-account: description

chassing:
Add a description to the `supportedDeploymentRegions` attribute.


TGPSKI:
AWS auto-enrolls accounts into 15 regions. Will we need to explicitly list the default list for every account?

PR #787, opened 2025-03-19
[AFL] Add optional title, description properties to metaschema

TGPSKI:
AFL - Automated Feature List
---

This MR adds optional title and description properties to `metaschema-1.json` child schemas. The MR also adds title and description pairs to `metaschema-1.json`, which are already defined in `/json-schema-spec-draft-06.json`.

The design doc did not include title field. I think it will be easier to source JSON/YAML property values versus depending on the filename and extracting the correct content.

### DIFF
--- a/schemas/metaschema-1.json
+++ b/schemas/metaschema-1.json
@@ -2,6 +2,8 @@
   "$schema": "/json-schema-spec-draft-06.json",
   "version": "1.0",
   "type": "object",
+  "title": "metaschema",
+  "description": "Defines the structure of a qontract-schema",
   "properties": {
     "$schema": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


TGPSKI:
When we get to later stages of the project, I'll move both fields to `required`. This means I'll be writing a lot of titles and descriptions, and taking plenty of notes on the way.

I'll also write and eventually enforce nested property parameters, something like this:

```json
...
      },
      "required": [
        "$schema"
      ],
      "patternProperties": {
        ".*": {
          "type": "object",
          "properties": {
            "description": {
              "type": "string"
            }
          },
          "required": [
            "description"
          ],
          "additionalProperties": true
        }
      }
    },
    "dependencies": {
      "type": "object"
    }
...
```

hemslo (approved):
LGTM!

chassing (approved):

PR #799, opened 2025-03-25
[AFL] query-validation, resource-template-test, template, template-collection, template-test

TGPSKI:
This MR adds titles and descriptions to query-validation, resource-template-test, template, template-collection, template-test schema.

### DIFF
--- a/schemas/app-interface/template-1.yml
+++ b/schemas/app-interface/template-1.yml
@@ -42,11 +42,17 @@ properties:
       path:
         type: string
         description: |
-          The path within the template to which the patch applies.
+          JSON path to the target field in the template to patch.
       identifier:
         type: string
         description: |
-          An identifier for the patch.
+          Unique identifier field to use with path results.
+          Can be a string (field name) or a JSON path.
+          Examples: 
+            path_results = [{'id': 1, 'name': 'a'}, {'id': 2, 'name': 'b'}]
+            identifier = 'id'
+            path_results = ['a', 'b']
+            identifier = '$.id'
     required:
     - path
   template:
--- a/schemas/app-interface/template-1.yml
+++ b/schemas/app-interface/template-1.yml
@@ -15,8 +15,6 @@ properties:
     type: string
     enum:
     - /app-interface/template-1.yml
-    description: |
-      The schema reference for this template file.
   name:
     type: string
     description: |
--- a/schemas/app-interface/template-1.yml
+++ b/schemas/app-interface/template-1.yml
@@ -1,9 +1,12 @@
 ---
 "$schema": /metaschema-1.json
 version: '1.0'
-
 type: object
-description: Template to be generated
+title: "App Interface Template"
+description: |
+  Schema for defining templates in app-interface. 
+  This includes details about the template, its target path, 
+  conditions, tests, and rendering options.
 
 additionalProperties: false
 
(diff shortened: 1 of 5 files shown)


chassing (approved):

schemas/app-interface/template-1.yml
```diff
@@ -12,43 +15,70 @@ properties:
...
   patch:
     type: object
+    description: |
+      A patch to be applied to the template.
     additionalProperties: false
     properties:
       path:
         type: string
+        description: |
+          The path within the template to which the patch applies.
```
hemslo, line 45: this is a json path to locate the section in target file to patch

schemas/app-interface/template-1.yml
```diff
@@ -12,43 +15,70 @@ properties:
...
       identifier:
         type: string
+        description: |
+          An identifier for the patch.
```
hemslo, line 49: this is an uniq identifier field to use on `path` results, can be a string as field name, or a json path string. For example if the `path` result is an array of object `[{'id': 1, 'name': 'a'}, {'id': 2, 'name': 'b'}]`, then identifier can be `id`. A special use case is `path` result is array of string `['a', 'b']`, then identifier can be the value itself, using json path `$`.

schemas/app-interface/template-test-1.yml
```diff
@@ -12,25 +15,33 @@ properties:
...
     type: string
-    description: the expected result of the templating
+    description: |
+      The expected output of the template after rendering.
   expectedToRender:
     type: boolean
-    description: whether the template is expected to render or not
-
+    description: |
+      Indicates whether the template is expected to render successfully.
```
hemslo, line 44: this is to test conditional render, when condition not met, we expect the template to not render

hemslo: to simplify the description, this is the expected evaluation result of `condition` in `schemas/app-interface/template-1.yml`.


Determines if the template should render based on the condition.
If true, the template should render ~~even if a required condition is not met~~.
If false, the template should not render ~~if a required condition is not met~~.

Then if true/false can be omitted together.

TGPSKI: I wrote out true/false because I had to re-read the description multiple times in order to comprehend it. I want to find a way to make descriptions immediately comprehensible.

Perhaps when I run into this situation, it means my attempt at description is not effective.

TGPSKI:
No idea what the failing Konflux test does

image

schemas/app-interface/template-test-1.yml
hemslo: maybe add the field link in description to make it clear

```
expected evaluation result of `$.condition` in `/app-interface/template-1.yml`
```

TGPSKI: Did you see my latest? 

`Indicates if template is expected to render given the test condition.`

TGPSKI: I'm lost.

template-test defines a test which validates the behavior of a template. Each template test should represent a test case with an expected outcome. Given a set of inputs to the template processor, we expect the output to look like {expected_output}. A template-test can test the generation of {expected_output}, or it can test the generation of {}.

If this parameter is true, then template-test output passes if {expected_output} returns. IF this parameter is false, then template-test passes if {} returns.

The test condition is defined by an instantiation of a test-template.

TGPSKI: TLDR: the boolean determines whether {expected_output} or {} is a passing test. "test condition" is the inputs provided to the template.

hemslo: no, that's not how it works, easier to show you the code, it's just verify if `condition` result matches `expectedToRender`.

https://github.com/app-sre/qontract-reconcile/blob/f772446c5acf6dc6147c002a9e033c370666e406/reconcile/templating/validator.py#L93-L106

TGPSKI: Coming back today with a fresh head =]

TGPSKI: I re-read the thread, looked at the code, and also checked examples of template and template-test. I still think we are saying the same thing. It's not important to keep the thread going.

TGPSKI: I'm just copy and pasting your desc from above and I'm moving on.

TGPSKI: `expectedConditionEvaluation` or similar makes more sense with description  `Expected evaluation result of `$.condition` in `/app-interface/template-1.yml`.`

PR #800, opened 2025-03-26
update OCM cluster storage quota list

patjlm:
New values appeared in OCM.

And we need higher values in some clusters


philipgough (approved):
/lgtm

PR #804, opened 2025-04-02
AFL: access/ external-user-1.yml, membership-provider-1.yml, oidc-permission-1.yml, permission-1.yml, role-1.yml, user-1.yml

TGPSKI:
This MR adds titles and descriptions to schemas/access/*.

### DIFF
--- a/schemas/access/permission-1.yml
+++ b/schemas/access/permission-1.yml
@@ -2,6 +2,12 @@
 "$schema": /metaschema-1.json
 version: "1.0"
 type: object
+title: "Permission"
+description: |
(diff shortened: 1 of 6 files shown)


schemas/access/permission-1.yml
```diff
@@ -105,10 +162,16 @@ oneOf:
       - github-org-team
     org:
       type: string
+      description: |
+        The GitHub organization associated with the team.
```
hemslo, line 166: can you check if `description` content need to be duplicated inside `oneOf`, or only define it once in root level.

PR #810, opened 2025-04-04
AFL: automated-action-argument-1

TGPSKI:
This pull request updates the schema file `schemas/app-sre/automated-action-argument-1.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/app-sre/automated-action-argument-1.yml
+++ b/schemas/app-sre/automated-action-argument-1.yml
@@ -2,6 +2,12 @@
 "$schema": /metaschema-1.json
 version: "1.0"
 type: object
+title: "Automated Action Argument"
+description: |
+  Schema for defining arguments for automated actions. 
+  Arguments specify the parameters required to execute an automated action, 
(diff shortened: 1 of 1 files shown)


schemas/app-sre/automated-action-argument-1.yml
```diff
@@ -13,30 +19,46 @@ properties:
...
+      The pattern used to match the name of resources in OpenShift. 
+      This is required when the type is `openshift`.
 
 oneOf:
 - properties:
     type:
       enum:
       - openshift
+      description: |
+        Specifies that the argument is related to OpenShift resources.
```
hemslo, line 48: similar question, do we need to duplicate all description inside `oneOf`

TGPSKI: https://github.com/app-sre/qontract-schemas/pull/867#discussion_r2169554767 thoughts?

PR #823, opened 2025-04-30
add quay username to external user

maorfr:
to be able to grant external users (IBM) access to RH controlled orgs


schemas/access/external-user-1.yml
```diff
@@ -15,15 +15,21 @@ properties:
...
+  roles:
+    type: array
+    items:
+      "$ref": "/common-1.json#/definitions/crossref"
+      "$schemaRef": "/access/role-1.yml"
 required:
 - $schema
 - labels
 - name
-- github_username
```
patjlm, line 28: this removal needs to be handled in qontract-reconcile: check if set

PR #824, opened 2025-05-01
fix(account-request): limit name pattern to match aws alias

hemslo:
Add pattern limit to aws account request name, to avoid error

```text
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateAccountAlias operation: The specified value for accountAlias is invalid. It must be a minimum length of 3 characters and maximum length of 63 characters, contain only digits, lowercase letters, and hyphens (-), but cannot begin or end with a hyphen. 
```

[APPSRE-11882](https://issues.redhat.com/browse/APPSRE-11882)

### DIFF
--- a/schemas/aws/account-request-1.yml
+++ b/schemas/aws/account-request-1.yml
@@ -13,6 +13,11 @@ properties:
     "$ref": "/common-1.json#/definitions/labels"
   name:
     type: string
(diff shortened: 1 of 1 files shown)


dwelch0 (approved):
LGTM

PR #831, opened 2025-05-20
SDE-4707 - Expand openshift-resources to support rhcs-cert provider

dwelch0:
Dependency for https://github.com/app-sre/qontract-reconcile/pull/5008


schemas/openshift/openshift-resource-1.yml
```diff
@@ -129,5 +137,27 @@ oneOf:
...
+      description: identifier for cert secret created in Vault and OpenShift namespace
+    service_account_name:
+      type: string
+      description: name of service account to utilize when requesting rhcsv2 certificate
+    service_account_password:
+      "$ref": "/common-1.json#/definitions/vaultSecret"
+    annotations:
+      "$ref": "/common-1.json#/definitions/annotations"
+    auto_renew_threshold_days:
+      type: integer
```
hemslo, line 157: `/dependencies/rhcs-cert-provider-1.yml` is not specified for each `rhcs-cert`, will there be multiple providers? Should we inline url and vault path here or put them into `app-interface-settings`?

dwelch0: From what I have gathered, RH IT currently only exposes a single url for requesting certs via web form but this may change.

Backing the url and vault path into `app-interface-settings` probably makes most sense currently. I will update

dwelch0: addressed in https://github.com/app-sre/qontract-schemas/pull/831/commits/1fc69ccc53551e3c30749636f0a5770fcf76b493

hemslo (approved):
LGTM!

PR #845, opened 2025-06-23
AFL: integration-1

TGPSKI:
This pull request updates the schema file `schemas/app-sre/integration-1.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/app-sre/integration-1.yml
+++ b/schemas/app-sre/integration-1.yml
@@ -2,6 +2,13 @@
 "$schema": /metaschema-1.json
 version: '1.0'
 type: object
+title: "Integration"
+description: |
+  Schema for defining an integration in app-interface. Integrations represent
+  automated processes or services that interact with app-interface data and
+  external systems. This schema allows specifying integration metadata, upstream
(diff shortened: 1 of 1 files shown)


schemas/app-sre/integration-1.yml
```diff
@@ -37,37 +56,65 @@ properties:
...
+        description: |
+          The command to execute for the PR check.
       state:
         type: boolean
+        description: |
+          Indicates if the PR check is stateful.
       sqs:
         type: boolean
+        description: |
+          Whether to send PR check results to SQS.
```
suzana-nesic, line 76: Maybe to keep consistency with the other descriptions, we can reword these descriptions to start with `If true,`

TGPSKI:
```suggestion
      state:
        type: boolean
        description: |
          If true, PR check job is stateful.
      sqs:
        type: boolean
        description: |
          If true, send PR check results to SQS.
```

patjlm (approved):

PR #847, opened 2025-06-23
AFL: integration-spec-1

TGPSKI:
This pull request updates the schema file `schemas/app-sre/integration-spec-1.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/app-sre/integration-spec-1.yml
+++ b/schemas/app-sre/integration-spec-1.yml
@@ -2,6 +2,13 @@
 "$schema": /metaschema-1.json
 version: "1.0"
 type: object
+title: "Integration Spec"
+description: |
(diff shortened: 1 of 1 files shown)


schemas/app-sre/integration-spec-1.yml
```diff
@@ -2,54 +2,85 @@
...
     type: string
-    description: command to run
     enum:
     - qontract-reconcile
     - app-interface-metrics-exporter
     - app-interface-reporter
     - glitchtip-access-reporter
     - glitchtip-access-revalidation
     - saas-metrics-exporter
+    description: |
```
suzana-nesic, line 35: For consistency, we can explain what true means here

PR #858, opened 2025-06-25
AFL: schedule-1

TGPSKI:
This pull request updates the schema file `schemas/app-sre/schedule-1.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/app-sre/schedule-1.yml
+++ b/schemas/app-sre/schedule-1.yml
@@ -2,6 +2,10 @@
 "$schema": /metaschema-1.json
 version: "1.0"
 type: object
+title: "Schedule"
+description: |
+  Schema for defining a schedule used to automate Slack user group assignments.
+  Schedules specify date ranges and the users assigned for each period.
 
 additionalProperties: false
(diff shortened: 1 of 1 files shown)


schemas/app-sre/schedule-1.yml
```diff
@@ -2,40 +2,65 @@
 "$schema": /metaschema-1.json
 version: "1.0"
 type: object
+title: "Schedule"
+description: |
+  Schema for defining a schedule used to automate Slack user group assignments.
```
patjlm, line 7: I'd mention this config is not related to slack but mostly used for slack user group assignment. We could reuse this schema for something else than slack.
```suggestion
  Schema for defining a schedule used to automate user group assignments. This is most notably used for Slack user groups via a permission file.
```

TGPSKI: Why not update the description when the use case changes, instead of implying a future use case that is currently unsupported?

PR #280, opened 2022-10-24
introduce service scorecard

maorfr:
part of https://issues.redhat.com/browse/APPSRE-6506

this PR adds the schemas as designed in https://gitlab.cee.redhat.com/service/sres-acceptance-criteria/-/merge_requests/10

PR #859, opened 2025-06-25
AFL: scorecard-2

TGPSKI:
This pull request updates the schema file `schemas/app-sre/scorecard-2.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/app-sre/scorecard-2.yml
+++ b/schemas/app-sre/scorecard-2.yml
@@ -2,6 +2,15 @@
 "$schema": /metaschema-1.json
 version: '1.0'
 type: object
+title: "Scorecard"
+description: |
+  Schema for defining a self-service scorecard for limited-scope service engagements.
+  The scorecard helps service owners assess their maturity and reliability
+  based on well-defined acceptance criteria established by AppSRE for
+  production services. Each scorecard is a self-assessment, to be filled
+  in and maintained by service owners, and is intended to be visible to
+  all stakeholders. The scorecard enables AppSRE to raise awareness and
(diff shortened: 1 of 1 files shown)


schemas/app-sre/scorecard-2.yml
```diff
@@ -2,6 +2,15 @@
 "$schema": /metaschema-1.json
 version: '1.0'
 type: object
+title: "Scorecard"
+description: |
+  Schema for defining a self-service scorecard for limited-scope service engagements.
```
patjlm, line 7:
```suggestion
  Schema for defining a service scorecard for limited-scope service engagements.
```

TGPSKI: This is funny to me! I reviewed all git history for every schema as I developed the documentation. The context for this schema is "a la carte" services vs fully fledged services. 

cc https://issues.redhat.com/browse/APPSRE-6506 "These scorecards are essentially self-assessments, to be **filled in by the service owners and to be kept up to date by them**.
Anyone at Red Hat should be able to see this maturity scorecard."

This schema seems purpose-built as alternative to SRE checkpoints.

TGPSKI: cc https://github.com/app-sre/qontract-schemas/pull/280 and https://github.com/app-sre/qontract-schemas/pull/287

patjlm: we will probably extend this a bit in the context of dedicated SRE. Let's review at that time

PR #867, opened 2025-06-25
AFL: account-1

TGPSKI:
This pull request updates the schema file `schemas/aws/account-1.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/aws/account-1.yml
+++ b/schemas/aws/account-1.yml
@@ -142,7 +142,7 @@ properties:
   enableDeletion:
     type: boolean
     description: |
-      If true, allows deletion of resources in this account.
+      If true, allows deletion of resources in this account without an explicit entry in 'deletionApprovals'.
 
   deletionApprovals:
     type: array
--- a/schemas/aws/account-1.yml
+++ b/schemas/aws/account-1.yml
@@ -2,6 +2,13 @@
 "$schema": /metaschema-1.json
 version: '1.0'
 type: object
+title: "AWS Account"
+description: |
+  Schema for defining an AWS account. This schema captures all metadata,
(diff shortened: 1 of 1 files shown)


patjlm (approved):

schemas/aws/account-1.yml
```diff
@@ -2,90 +2,166 @@
...
+    description: |
+      The default AWS region for resources created in this account.
+
   supportedDeploymentRegions:
     type: array
     items:
       type: string
-      description: "The regions where the account is enabled and can be used."
+      description: |
+        The regions where the account is enabled and can be used for
```
rporres, line 56:
```suggestion
        The region where the account is enabled and can be used for
```

schemas/aws/account-1.yml
```diff
@@ -2,90 +2,166 @@
...
+          The phone number of the security contact.
     required:
     - name
     - email
     - phoneNumber
+
   automationToken:
     "$ref": "/common-1.json#/definitions/vaultSecret"
+    description: |
+      Reference to the Vault secret containing the automation token for
```
rporres, line 123:
```suggestion
      Reference to the Vault secret containing the credentials for
```

TGPSKI: I admit I'm confused with these suggestions. Why add descriptions with details that do not match the property name?

schemas/aws/account-1.yml
TGPSKI: This is intentionally plural. The property is `supportedDeploymentRegions`, not `supportedDeploymentRegion`. This array can support multiple regions.

PR #882, opened 2025-06-28
remove non-existing fields from graphql schema.yml

patjlm:
These field have no equivalent in schema files. They're not queried in qontract-reconcile.


TGPSKI (approved):
LGTM

patjlm:
Added other fixes in https://github.com/app-sre/qontract-schemas/pull/882/commits/6a216250d4e6a6b72d4b4c79b7cac6c8e6578597

PR #891, opened 2025-06-28
AFL: dns-record-1

TGPSKI:
This pull request updates the schema file `schemas/dependencies/dns-record-1.yml` to include a title, description for the top-level schema, and descriptions for all properties. All descriptions use multiline YAML syntax.

### DIFF
--- a/schemas/dependencies/dns-record-1.yml
+++ b/schemas/dependencies/dns-record-1.yml
@@ -2,6 +2,13 @@
 "$schema": /metaschema-1.json
 version: '1.0'
 type: object
+title: "DNS Record"
+description: |
+  Schema for referencing a DNS record as a dependency. This schema enables
+  app-interface to declare DNS records for integration, documentation, and
+  automation purposes. It does not provision or manage the lifecycle of the
+  DNS record itself. Supports standard and advanced DNS record types, routing
+  policies, health checks, and integration with OpenShift clusters and namespaces.
 
 additionalProperties: false
 properties:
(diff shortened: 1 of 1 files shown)


schemas/dependencies/dns-record-1.yml
```diff
@@ -2,15 +2,27 @@
 "$schema": /metaschema-1.json
 version: '1.0'
 type: object
+title: "DNS Record"
+description: |
+  Schema for referencing a DNS record as a dependency. This schema enables
+  app-interface to declare DNS records for integration, documentation, and
+  automation purposes. It does not provision or manage the lifecycle of the
+  DNS record itself. Supports standard and advanced DNS record types, routing
```
patjlm, line 10: ref
> This schema enables app-interface to declare DNS records for integration, documentation, and automation purposes. It does not provision or manage the lifecycle of the DNS record itself.

It does. The records will get created/updated/deleted. I am not sure what you mean here. Do you mean it's not a top-level queriable object?

TGPSKI: I understood all dependency/* schemas as references to existing declarations, not declarations themselves. Perhaps that is incorrect, or there are conflicting patterns in the codebase. 

**The pattern that feels correct to me:**

`./schemas/cloudflare/dns-record-1 and dns-zone-1` declare resources and we expect them to exist after declaration.

`./schemas/dependencies/dns-record-1 and dns-zone-1` marks a dependency relationship with an existing resource.

**What I'm learning:**

`./schemas/dependencies/dns-record-1 and dns-zone-1` are AWS specific resource declarations. 

**Larger suggestion:**

* AWS-specific schemas that declare resources should live in ./schemas/aws
* Dependency relationships between existing resources should be declared with `./schemas/dependencies/*`

PR #909, opened 2025-07-31
Enable VPC and Subnets tagging in the VPC management

lechuk47:
- **Add VPC request tags**
- **Add VPC and subnets tags**

Ref: https://github.com/app-sre/qontract-reconcile/pull/5150

### DIFF
--- a/schemas/aws/vpc-request-1.yml
+++ b/schemas/aws/vpc-request-1.yml
@@ -51,7 +51,7 @@ properties:
       public_subnet_tags:
         description: Public subnets tags.
         type: object
-  vpc_additional_tags:
+  vpc_tags:
     description: Additional tags for the VPC.
     type: object
 
--- a/schemas/aws/vpc-request-1.yml
+++ b/schemas/aws/vpc-request-1.yml
@@ -45,6 +45,16 @@ properties:
         type: array
         items:
           type: string
+      private_subnet_tags:
+        description: Private subnets tags.
+        type: object
+      public_subnet_tags:
+        description: Public subnets tags.
+        type: object
+  vpc_additional_tags:
+    description: Additional tags for the VPC.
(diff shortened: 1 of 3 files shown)


schemas/aws/vpc-request-1.yml
```diff
@@ -45,6 +45,16 @@ properties:
         type: array
         items:
           type: string
+      private_subnet_tags:
+        description: Private subnets tags.
+        type: object
+      public_subnet_tags:
+        description: Public subnets tags.
+        type: object
```
hemslo, line 53: do you want to keep these 2 in root level to be consistent with `schemas/aws/vpc-1.yml` file, or move vpc-1.yml ones to be under `subnets`

hemslo: is there a use case that each subnet to have different tags? we can even add `tags` to subnet object

lechuk47: Yes, 

These tags are used to auto-discover subnets, and there are cases where subnets (or AZs) need to be excluded. For example, RDS Proxy is not supported in us-east-1e 

[Ref](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-proxy.html#rds-proxy.RegionVersionAvailability): 
```
Certain Regions have Availability-Zone (AZ) restrictions to consider while creating your proxy. US East (N. Virginia) Region does not support RDS Proxy in the use1-az3 Availability Zone. US West (N. California) Region does not support RDS Proxy in the usw1-az2 Availability Zone. When selecting subnets while creating your proxy, make sure that you don't select subnets in the Availability Zones mentioned above.
```

The tf module also allows setting tags per az with `private_subnet_tags_per_az`

hemslo (approved):
LGTM!

PR #910, opened 2025-08-01
SDE-4648 - Add optional attributes for slo alert generation

dwelch0:
Dependency for https://github.com/app-sre/qontract-reconcile/pull/5153

These optional attributes support sloth alert generation: https://sloth.dev/examples/default/getting-started/


bkez322 (approved):
LGTM

Nice detailed descriptions.

PR #922, opened 2025-08-26
yamllint

chassing:
Introduce `yamllint` checks and lint all files.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -41,6 +41,6 @@ rules:
   key-duplicates: {}
   new-line-at-end-of-file: {}
   quoted-strings:
-    quote-type: double
+    quote-type: any
     required: false
   trailing-spaces: {}
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,46 @@
+---
+locale: en_US.UTF-8
+
+ignore:
+- ".tox"
+
+rules:
+  braces:
+    min-spaces-inside: 0
+    max-spaces-inside: 1
+    min-spaces-inside-empty: 0
+    max-spaces-inside-empty: 0
+  brackets:
+    forbid: non-empty
+  colons:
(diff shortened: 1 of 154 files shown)


mehfuz (approved):
LGTM ....

.yamllint
```diff
@@ -0,0 +1,46 @@
...
+  hyphens:
+    max-spaces-after: 1
+  indentation:
+    indent-sequences: false
+    spaces: 2
+  line-length: disable
+  key-duplicates: {}
+  new-line-at-end-of-file: {}
+  quoted-strings:
+    quote-type: double
```
hemslo, line 44: double quote allows escape and single quote just literal, should we use `any` to avoid `# yamllint disable-line rule:quoted-strings`?

Dockerfile
```diff
@@ -1,23 +1,22 @@
...
     update-alternatives --install /usr/bin/python3 python /usr/bin/python3.11 1 && \
     ln -snf /usr/bin/python3.11 /usr/bin/python && \
     python3 -m ensurepip
 
 RUN python3 -m pip install --no-cache-dir --upgrade pip tox
 
-COPY tox.ini test /schemas/
+COPY tox.ini test .yamllint /schemas/
 
 CMD ["tox"]
```
hemslo, line 22: we have `tox` in this repo? Set it in `CMD` won't actually run it in pipeline. 

We don't need `test` at all, can just use qontract validator to do bundle validation. We need to get rid of tox and reimplement `make test bundle validate gql_validate` in Dockerfile.

chassing: Good idea. I'll tackle that in a future PR. https://issues.redhat.com/browse/APPSRE-12422

hemslo (approved):
LGTM!

PR #924, opened 2025-08-29
feat: :lock: add schema definitions for disabling IAM keys

cubismod:
https://issues.redhat.com/browse/PROJQUAY-9354

Adding support to `aws-iam-keys` integration for disabling keys, in addition to deleting them for use cases where we aren't using the access keys.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -1356,6 +1356,7 @@ confs:
   - { name: deletionApprovals, type: DeletionApproval_v1, isList: true }
   - { name: disable, type: DisableClusterAutomations_v1 }
   - { name: deleteKeys, type: string, isList: true }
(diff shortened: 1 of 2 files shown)


graphql-schemas/schema.yml
```diff
@@ -1356,6 +1356,7 @@ confs:
   - { name: deletionApprovals, type: DeletionApproval_v1, isList: true }
   - { name: disable, type: DisableClusterAutomations_v1 }
   - { name: deleteKeys, type: string, isList: true }
+  - { name: disableKeys, type: string, isList: true }
```
esron, line 1359: nitpick: if it is a list it should be `disabledKeys`

But I see that `deleteKeys` is already a thing so this is far from a blocker.

PR #936, opened 2025-10-07
Add secret_format to AWS Secrets Manager

rporres:
The idea is to be able to create templated secrets to copy from Vault to AWS Secrets Manager

This work is needed in the context of APPSRE-12414, since AWS expects the RDS credentials for AWS RDS Proxy in a defined format that cannot be configured.


chassing (approved):

chassing:
If you merge and promote. Could you please also take #816 with you? 🙇

PR #994, opened 2026-04-24
introduce vpc-endpoint and vpc-endpoint-service providers to terraform-resource-2.yml

jfchevrette:
Registers two new ERv2 providers that implement AWS PrivateLink connectivity between OpenShift clusters

## Changes

Adds two new provider entries to schemas/aws/terraform-resource-2.yml:

- `vpc-endpoint-service` — provider side. Creates an AWS VPC Endpoint Service attached to an existing NLB provisioned by an OpenShift Service of type LoadBalancer. The NLB is discovered automatically via kubernetes.io/service-name tags.
- `vpc-endpoint` — consumer side. Creates an interface VPC Endpoint connecting to a vpc-endpoint-service resource. Handles private subnet selection and AZ alignment automatically.

## Related

  - Module repos: [er-aws-vpc-endpoint-service](https://github.com/app-sre/er-aws-vpc-endpoint-service), [er-aws-vpc-endpoint](https://github.com/app-sre/er-aws-vpc-endpoint)
  - [Design doc](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/app-sre/design-docs/er-aws-privatelink.md)

### DIFF
--- a/schemas/aws/terraform-resource-2.yml
+++ b/schemas/aws/terraform-resource-2.yml
@@ -259,6 +259,17 @@ properties:
       type: string
   openshift_ingress_load_balancer_arn:
     type: string
+  openshift_service_name:
+    type: string
+  allowed_consumer_clusters:
+    type: array
+    items:
+      "$ref": "/common-1.json#/definitions/crossref"
+      "$schemaRef": "/openshift/cluster-1.yml"
+  allowed_principal_arns:
+    type: array
+    items:
+      type: string
   insights_callback_urls:
     type: array
     items:
@@ -2058,5 +2069,46 @@ oneOf:
   required:
   - identifier
   - module
+- additionalProperties: false
+  properties:
+    provider:
+      type: string
+      enum:
+      - vpc-endpoint-service
+    identifier:
+      "$ref": "/common-1.json#/definitions/longIdentifier"
+    region:
+      "$ref": "/aws/regions-1.yml#/properties/region"
+    openshift_service_name:
+      type: string
+      description: Name of the OpenShift Service (type LoadBalancer) in the target namespace whose NLB is exposed via the VPC Endpoint Service
+    allowed_consumer_clusters:
+      type: array
+      description: Clusters allowed to connect to the VPC Endpoint Service. Each cluster's account uid and terraformUsername are used to build an IAM principal ARN of the form arn:aws:iam::<uid>:user/<terraformUsername>
+      items:
+        "$ref": "/common-1.json#/definitions/crossref"
+        "$schemaRef": "/openshift/cluster-1.yml"
+    allowed_principal_arns:
+      type: array
+      description: Explicit IAM principal ARNs allowed to connect to the VPC Endpoint Service
+      items:
+        type: string
+    output_resource_name:
+      "$ref": "/common-1.json#/definitions/longIdentifier"
+    tags:
+      "$ref": "/common-1.json#/definitions/labels"
+    annotations:
+      "$ref": "/common-1.json#/definitions/annotations"
+    managed_by_erv2:
+      type: boolean
+      description: Manage the resource with ERv2
+    delete:
+      type: boolean
+      description: Flag a resource to be deleted
+    module_overrides:
+      "$ref": "/external-resources/module-overrides-1.yml"
+  required:
+  - identifier
+  - openshift_service_name
 required:
 - provider
--- a/schemas/aws/terraform-resource-2.yml
+++ b/schemas/aws/terraform-resource-2.yml
@@ -270,6 +270,8 @@ properties:
     type: array
     items:
       type: string
+  endpoint_service_name:
+    type: string
   insights_callback_urls:
     type: array
     items:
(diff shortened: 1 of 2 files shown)


schemas/aws/terraform-resource-2.yml
```diff
@@ -2058,5 +2067,70 @@ oneOf:
...
+    vpc:
+      "$ref": "/common-1.json#/definitions/crossref"
+      "$schemaRef": "/aws/vpc-1.yml"
+    output_resource_name:
+      "$ref": "/common-1.json#/definitions/longIdentifier"
+    tags:
+      "$ref": "/common-1.json#/definitions/labels"
+    annotations:
+      "$ref": "/common-1.json#/definitions/annotations"
+    managed_by_erv2:
```
fishi0x01, line 2136: Im not 100% sure, but will there be a migration? Historically we used this flag to mark already migrated resources. Im not sure this is needed if we dont have anything to migrate. Iirc the plan was to remove those flags from resources once migrations finished.

schemas/aws/terraform-resource-2.yml
```diff
@@ -2058,5 +2067,70 @@ oneOf:
...
+- additionalProperties: false
+  properties:
+    provider:
+      type: string
+      enum:
+      - vpc-endpoint
+    identifier:
+      "$ref": "/common-1.json#/definitions/longIdentifier"
+    endpoint_service_name:
+      type: string
```
fishi0x01, line 2124: where does this point to? Trying to figure out if enum or crossref (if this is supposed to point to some def in our schema) do make any sense here rather than free string.

Maybe do you have an example of what one would put here?

schemas/aws/terraform-resource-2.yml
```diff
@@ -2058,5 +2067,70 @@ oneOf:
...
+      items:
+        "$ref": "/common-1.json#/definitions/crossref"
+        "$schemaRef": "/openshift/cluster-1.yml"
+    output_resource_name:
+      "$ref": "/common-1.json#/definitions/longIdentifier"
+    tags:
+      "$ref": "/common-1.json#/definitions/labels"
+    annotations:
+      "$ref": "/common-1.json#/definitions/annotations"
+    managed_by_erv2:
```
fishi0x01, line 2104: same - not sure it is needed

schemas/aws/terraform-resource-2.yml
```diff
@@ -2058,5 +2067,70 @@ oneOf:
...
+    provider:
+      type: string
+      enum:
+      - vpc-endpoint-service
+    identifier:
+      "$ref": "/common-1.json#/definitions/longIdentifier"
+    region:
+      "$ref": "/aws/regions-1.yml#/properties/region"
+    openshift_service_name:
+      type: string
```
fishi0x01, line 2085: same, do you have a concrete example for an outsider what this is supposed to point to exactly? Maybe some doc string for that field?

fishi0x01 (commented):
Awesome! ❤️ 
Couple of questions that stem from me not knowing enough about this module (yet) :)

schemas/aws/terraform-resource-2.yml
hemslo: this is still needed until all ERv2 resources migrated

schemas/aws/terraform-resource-2.yml
```diff
@@ -2058,5 +2067,70 @@ oneOf:
...
+    allowed_consumer_clusters:
+      type: array
+      items:
+        "$ref": "/common-1.json#/definitions/crossref"
+        "$schemaRef": "/openshift/cluster-1.yml"
```
hemslo, line 2092: We can get AWS account info from ROSA cluster to be used here, how about OSD clusters, those cluster spec has no `spec.account`.

schemas/aws/terraform-resource-2.yml
fishi0x01: Not sure I understand this fully. We do not have any privatelink tf resources right? Or you mean until we decommission tf-resources completely (regardless of resource type)?

hemslo: yes, need to completely get rid of terraform resources

fishi0x01: Ok makes sense then. Thank you for the explanation :+1:

schemas/aws/terraform-resource-2.yml
jfchevrette: I excluded OSD from this first MVP. This is being used to build the ARNs that are allowed to link to the VPCES. For OSD we'll have to change this a bit for sure

schemas/aws/terraform-resource-2.yml
jfchevrette: There are docs in the related app-interface MR

schemas/aws/terraform-resource-2.yml
jfchevrette: There are docs in the related app-interface MR

hemslo: please add `description` for new fields

schemas/aws/terraform-resource-2.yml
jfchevrette: Looks like that field is expected to exist per https://github.com/app-sre/qontract-reconcile/blob/120e07a3f59295eb11fb4fbb2d26e18176488a3f/reconcile/external_resources/model.py#L131

schemas/aws/terraform-resource-2.yml
jfchevrette: Added description fields

schemas/aws/terraform-resource-2.yml
jfchevrette: Added description fields

schemas/aws/terraform-resource-2.yml
jfchevrette: Leaving the field `managed_by_erv2` as it is being checked throughout the external-resources. A refactor of that could be needed if we want to allow not setting the field for resources which are erv2-only

fishi0x01 (approved):
LGTM!

hemslo (approved):
LGTM!

PR #999, opened 2026-05-01
long-standing TODO item for comment improvement

lisa:
In #954, [Di pointed out](https://github.com/app-sre/qontract-schemas/pull/954#discussion_r2570228631) that the previous documentation string for this attribute was not correctly describing what the attribute was for/doing. I told him that I would address the issue, but the end of 2025 hit and then in the new year, I had forgotten.

### DIFF
--- a/schemas/openshift/cluster-1.yml
+++ b/schemas/openshift/cluster-1.yml
@@ -34,11 +34,11 @@ properties:
   allowedToBypassPublicPeeringRestriction:
     type: boolean
     description: |
-      In normal operation, peering with a public cluster is not
-      allowed for security reasons. In very rare cases we need an
-      exception for the prohibition. Use this option with extreme
-      caution and after consulting the appropriate security
-      professionals.
+      In normal operation, peering a pair of public clusters is not allowed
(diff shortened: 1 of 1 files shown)


schemas/openshift/cluster-1.yml
```diff
@@ -34,11 +34,11 @@ properties:
...
-      In normal operation, peering with a public cluster is not
-      allowed for security reasons. In very rare cases we need an
-      exception for the prohibition. Use this option with extreme
-      caution and after consulting the appropriate security
-      professionals.
+      In normal operation, peering a pair of public clusters is not allowed
+      (they could simply communicate over the Internet). In very rare cases we
+      need an exception from the prohibition to gain more performance security
+      for the interesting traffic. Use this option with extreme caution and
+      after consulting the appropriate security professionals.
```
hemslo: It's more a cost / aws quota limit concern rather than security, as enable it has better performance & security.

lisa: That's fair! I will update.

hemslo (approved):
LGTM!

PR #1000, opened 2026-05-04
chore(glitchtip): remove unused ldapGroups and membersOrganizationRol…

jfreixa:
…e fields

Follow-up to app-sre/qontract-reconcile#5516 which removed all LDAP resolution logic from the glitchtip-api integration. No glitchtip team in app-interface uses these fields, making them dead schema.


hemslo (approved):
LGTM!

PR #1010, opened 2026-05-13
add merge_limit and consecutive_failure_limit to housekeeping s…

rhur-pixel:
### Summary
Add optional `merge_limit` and `consecutive_failure_limit` integer fields to the `gitlabHousekeeping` configuration in both the JSON schema and GraphQL schema.
- `merge_limit`: controls how many MRs can be merged per loop iteration (defaults to `limit` when absent)
- `consecutive_failure_limit`: controls how many consecutive pipeline failures before an MR is flagged (defaults to 3 when absent)
Both fields are optional and backward compatible with existing configs.
### Testing
- `make bundle-and-validate-schema` -- passed
- Bundled against local app-interface data with `qontract-bundler` -- passed
- `qontract-validator --only-errors` -- `[]` (no errors)
### Ticket
[APPSRE-13997](https://redhat.atlassian.net/browse/APPSRE-13997)

### DIFF
--- a/schemas/app-sre/app-1.yml
+++ b/schemas/app-sre/app-1.yml
@@ -292,6 +292,16 @@ properties:
               not:
                 enum:
                 - 0
+            merge_limit:
+              type: integer
+              not:
+                enum:
+                - 0
+            consecutive_failure_limit:
+              type: integer
+              not:
+                enum:
+                - 0
             enable_closing:
               type: boolean
             pipeline_timeout:
--- a/schemas/app-sre/app-1.yml
+++ b/schemas/app-sre/app-1.yml
@@ -294,14 +294,12 @@ properties:
                 - 0
             merge_limit:
               type: integer
-              not:
-                enum:
-                - 0
+              minimum: 1
+              description: "Maximum MRs merged per loop iteration. Defaults to limit if unset."
             consecutive_failure_limit:
               type: integer
-              not:
-                enum:
-                - 0
+              minimum: 1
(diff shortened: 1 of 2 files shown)


schemas/app-sre/app-1.yml
```diff
@@ -292,6 +292,16 @@ properties:
               not:
                 enum:
                 - 0
+            merge_limit:
+              type: integer
+              not:
+                enum:
+                - 0
```
TGPSKI:
```suggestion
              "$ref": "/common-1.json#/definitions/positiveInteger"
              description: add a description
```

schemas/app-sre/app-1.yml
```diff
@@ -292,6 +292,16 @@ properties:
...
+            consecutive_failure_limit:
+              type: integer
+              not:
+                enum:
+                - 0
```
TGPSKI:
```suggestion
              "$ref": "/common-1.json#/definitions/positiveInteger"
              description: add a description
```

TGPSKI (changes requested):

schemas/app-sre/app-1.yml
rhur-pixel: The description is a good call I'll add that, but the ref for positiveInteger has type: number, not integer. I was more or less trying to match the limit and days_interval pattern. $ref seems cleaner but I was just wondering if the positiveInteger definition might cause problems.

TGPSKI (approved):

PR #1026, opened 2026-06-24
fix(schema): make ai- prefix optional for ldapGroup names

dmzoneill:
## Summary

- Make the `ai-` prefix optional in the `ldapGroupName` pattern in `schemas/common-1.json`
- Changes pattern from `^ai-[a-z][a-z0-9-_.]+[a-z0-9]*$` to `^(ai-)?[a-z][a-z0-9-_.]+[a-z0-9]*$`

## Why

Many existing Rover groups do not follow the `ai-` naming convention. This prevents teams from linking them via `ldapGroup` in app-interface roles. For example:

- `automation-analytics`
- `dataverse-source-aapautomationanalytics`
- `automation-analytics-sonarqube-users`

All of these are valid Rover groups but fail the current schema validation.

## Impact

- Existing `ai-` prefixed groups continue to work unchanged
- Groups without the `ai-` prefix can now be referenced

## Jira

AAP-80318

### DIFF
--- a/schemas/common-1.json
+++ b/schemas/common-1.json
@@ -191,7 +191,7 @@
     },
     "ldapGroupName": {
       "type": "string",
-      "pattern": "^ai-[a-z][a-z0-9-_.]+[a-z0-9]*$"
+      "pattern": "^(ai-)?[a-z][a-z0-9-_.]+[a-z0-9]*$"
     },
     "positiveInteger": {
       "type": "number",


chassing:
Unfortunately, we can't do that. The `ai-` prefix is mandatory and enforced by the IT. Our AppSRE service account doesn't have permission to manage other than `ai-` prefixed LDAP groups.

PR #1028, opened 2026-06-25
add jwt to the auth methods available for Vault

jonmosco:
1. Added jwt to the type enum
2. Added a JWT config block to settings.config.oneOf supporting three validation methods (mutually exclusive per Vault docs): oidc_discovery_url, jwks_url, or jwt_validation_pubkeys, plus optional default_role and bound_issuer

### DIFF
--- a/schemas/vault-config/auth-1.yml
+++ b/schemas/vault-config/auth-1.yml
@@ -36,12 +36,14 @@ properties:
     enum:
     - approle
     - github
+    - jwt
     - oidc
     - kubernetes
     description: |
       The type of Vault auth method to enable. Supported types include:
         - 'approle': Machine-to-machine authentication using AppRole IDs and secrets.
         - 'github': User authentication via GitHub organizations and teams.
+        - 'jwt': Machine or user authentication via JSON Web Token validation.
         - 'oidc': User authentication via OpenID Connect providers.
         - 'kubernetes': Service account authentication from Kubernetes clusters.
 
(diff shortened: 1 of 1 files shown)


a-sinclaire:
LGTM

RH-tj:
## Suggested hardening for the JWT auth config block

See linked patch with four changes (2 high, 2 medium priority) that tighten the schema to better align with Vault's API surface and IETF security guidance.

**Full patch:** https://gist.github.com/RH-tj/4230fb225a8717ceb21070fa30772607

### HIGH: Add `jwt_supported_algs` field

Vault defaults to accepting **all** signing algorithms for JWT roles when this field is unset. This violates [RFC 8725](https://www.rfc-editor.org/rfc/rfc8725) (JWT Best Current Practices), which mandates explicit algorithm whitelisting. The patch adds an array field with an enum of Vault-supported algorithms, excluding `none`.

### HIGH: Enforce mutual exclusivity via `oneOf`

Vault's API docs state: *"One (and only one) of `oidc_discovery_url`, `jwks_url`, and `jwt_validation_pubkeys` must be set."* The current PR documents this in description text but doesn't enforce it at the schema level. The patch adds a `oneOf` constraint with explicit `required` blocks so invalid configurations are caught at validation time, not at Vault apply time.

### MEDIUM: Add `jwks_ca_pem` and `oidc_discovery_ca_pem` fields

Vault supports CA certificate pinning for TLS connections to JWKS and OIDC discovery endpoints. Without these fields, operators in internal PKI environments must either bypass TLS verification or manage CA trust outside the schema. The patch adds both as optional string fields.

### MEDIUM: Add RFC references to field descriptions

Field descriptions currently use informal prose with no traceability to the standards they implement. The patch adds references to RFC 7515 (JWS), RFC 7517 (JWK), RFC 7518 (JWA), RFC 7519 (JWT), RFC 8725 (JWT BCP), and OpenID Connect Discovery 1.0 where applicable.

---

The patch is purely additive — it does not remove or alter any existing fields from the PR. Happy to discuss or adjust.

RH-tj (changes requested):
see comment and gist: https://gist.github.com/RH-tj/4230fb225a8717ceb21070fa30772607

jonmosco:
Regarding https://github.com/app-sre/qontract-schemas/pull/1028#issuecomment-4802842589:

Thanks for the review. For this PR, I’m trying to keep the schema intentionally minimal — basically just enough to accept the JWT auth type and let vault-manager configure it.

Vault already validates things like supported algorithms, mutually exclusive settings, and TLS config when the config is applied, so I don’t think we gain much by duplicating all of that logic in the schema right now. It also adds maintenance overhead, especially since JWT/OIDC behavior can evolve and hardcoded enums can get stale.

RH-tj (approved):
approved after scope clarification from @jonmosco

PR #1048, opened 2026-08-07
feat(acs): add optional instance field to ACS schemas

casey-williams-rh:
## Overview

> **Ticket:** [APPSRE-14643](https://redhat.atlassian.net/browse/APPSRE-14643)
> **Phase:** 1 of 4

This MR adds an optional `instance` field to the ACS permission and policy schemas. The field links a permission or policy to a specific ACS instance.

The field is optional in this MR. A follow-up MR (Phase 4) makes it required after the app-interface data migration.

## Why

The `acs-rbac` and `acs-policies` integrations support only one ACS instance. The code raises an error when more than one `acs-instance-1.yml` datafile exists. This blocks onboarding of new ACS instances. Adding the `instance` field is the first step to support multiple instances.

#### `schemas/access/oidc-permission-1.yml`

Adds an optional `instance` crossref to the ACS `oneOf` branch. The ref points to `/dependencies/acs-instance-1.yml`. The existing ACS `required` arrays (`permission_set + clusters` and `permission_set + namespaces`) do not change.

#### `schemas/acs/acs-policy-1.yml`

Adds an optional `instance` crossref property. The ref points to `/dependencies/acs-instance-1.yml`. The top-level `required` array does not change.

#### `graphql-schemas/schema.yml`

Adds an optional `instance: AcsInstance_v1` field to two types:

- `AcsPolicy_v1`
- `OidcPermissionAcs_v1`

## Merge order

Merge this MR before the qontract-reconcile MR. The qontract-reconcile CI pipeline validates against these schemas.

1. **This MR** — schemas (instance optional)
2. **qontract-reconcile MR** — code reads and uses the instance field
3. **app-interface MR** — tags all existing permissions and policies with `instance: app-sre-acs`
4. **Follow-up schema MR** — makes `instance` required after Phase 3 merges

## Validation

```bash
GIT_COMMIT=$(git rev-parse HEAD) GIT_COMMIT_TIMESTAMP=$(git log -1 --format=%ct) \
make bundle-and-validate-schema
```

A passing run prints `[]`.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -751,6 +751,7 @@ confs:
   - { name: path, type: string, isRequired: true }
   - { name: labels, type: json }
   - { name: name, type: string, isRequired: true }
+  - { name: instance, type: AcsInstance_v1 }
   - { name: description, type: string }
   - { name: integrations, type: AcsPolicyIntegrations_v1 }
   - { name: severity, type: string, isRequired: true }
@@ -3398,6 +3399,7 @@ confs:
   - { name: name, type: string, isRequired: true, isUnique: true }
   - { name: service, type: string, isRequired: true }
   - { name: description, type: string, isRequired: true }
+  - { name: instance, type: AcsInstance_v1 }
   - { name: permission_set, type: string, isRequired: true }
   - { name: clusters, type: Cluster_v1, isList: true }
   - { name: namespaces, type: Namespace_v1, isList: true }
--- a/schemas/access/oidc-permission-1.yml
+++ b/schemas/access/oidc-permission-1.yml
@@ -73,6 +73,9 @@ oneOf:
     service:
       enum:
       - acs
+    instance:
+      "$ref": "/common-1.json#/definitions/crossref"
+      "$schemaRef": "/dependencies/acs-instance-1.yml"
     clusters:
       type: array
       items:
--- a/schemas/acs/acs-policy-1.yml
+++ b/schemas/acs/acs-policy-1.yml
@@ -25,6 +25,11 @@ properties:
     type: string
     description: |
       The name of the policy. This should be unique and descriptive.
+  instance:
+    "$ref": "/common-1.json#/definitions/crossref"
+    "$schemaRef": "/dependencies/acs-instance-1.yml"
+    description: |
+      Reference to the ACS instance this policy applies to.
   integrations:
     type: object
     description: |


suzana-nesic:
Is there ever a time where one oidc permission or ACS policy file will need multiple instances defined or is the expected pattern a 1:1?

casey-williams-rh:
I think the expected pattern is 1:1.

For OIDC permissions, a permission file defines two things: a `permission_set` and a list of `clusters` or `namespaces`. The reconciler uses this file to create one role in a single ACS instance.

As I understand it, each ACS instance keeps its own list of clusters. When a cluster joins an instance, that instance gives it a private internal ID. `Instance B` does not know about the clusters in `Instance A`.

So, a permission that lists `cluster: app-sre-prod` only works on the instance where `app-sre-prod` is registered. You can't use one permission for two instances, since each instance has its own cluster list.

I think it's the same for ACS policies. A policy scope is a list of `clusters` or `namespaces`. The reconciler calls `acs_api.list_clusters()` to get the cluster IDs, but that call only returns clusters registered with the current instance.

You can't apply one policy file to two instances, because each one uses different cluster IDs.

suzana-nesic (approved):

suzana-nesic:
> I think the expected pattern is 1:1.
> 
> For OIDC permissions, a permission file defines two things: a `permission_set` and a list of `clusters` or `namespaces`. The reconciler uses this file to create one role in a single ACS instance.
> 
> As I understand it, each ACS instance keeps its own list of clusters. When a cluster joins an instance, that instance gives it a private internal ID. `Instance B` does not know about the clusters in `Instance A`.
> 
> So, a permission that lists `cluster: app-sre-prod` only works on the instance where `app-sre-prod` is registered. You can't use one permission for two instances, since each instance has its own cluster list.
> 
> I think it's the same for ACS policies. A policy scope is a list of `clusters` or `namespaces`. The reconciler calls `acs_api.list_clusters()` to get the cluster IDs, but that call only returns clusters registered with the current instance.
> 
> You can't apply one policy file to two instances, because each one uses different cluster IDs.

Thank you for the explanation! What you have is good. I'm not sure if at any point you'll want to change the schema as you want the q-r portion but I went ahead and signed off on what you currently have. Is you need to update this PR for any reason, let us know for another review

hemslo (commented):
Before we add more features to ACS related integrations, we should spike if these should still be managed by qontract-reconcile via acs api, as now ACS has native as code way https://developers.redhat.com/articles/2026/03/24/integrate-red-hat-advanced-cluster-management-argo-cd

bkez322:
> Before we add more features to ACS related integrations, we should spike if these should still be managed by qontract-reconcile via acs api, as now ACS has native as code way https://developers.redhat.com/articles/2026/03/24/integrate-red-hat-advanced-cluster-management-argo-cd

@hemslo this doc covers how to use argocd to manage acs-cluster-registration, which is probably a good idea for the future

I can't find any clean way of managing ACS-RBAC via any existing argo-native solutions. In order to do that, I think we'd need to build our own brand-new acs-rbac kubernetes-controller.

For the problem we're trying to solve (manage RBAC for multiple ACS-instances), I think this QR enhancement is our least bad option today.

hemslo:
https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_security_for_kubernetes/4.11/html/configuring/declarative-configuration-using can be used to manage rbac as code. We can do quick changes to support multiple instances in qontract schemas and integrations for https://redhat.atlassian.net/browse/APPSRE-14643, and replace them with native solutions later.

bkez322:
>https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_security_for_kubernetes/4.11/html/configuring/declarative-configuration-using can be used to manage rbac as code.

@hemslo I think there might be an expectation disconnect here.

This doc explains managing RBAC for RHACS- i.e. an ACS Central that we would be fully managing the lifecycle of on our Openshift cluster.

https://acs-data-cl4gmo2o5v7imos30r30.acs.rhcloud.com/ is an ACSCS Central instance that is managed by ACSCS and we basically just consume it like a saas service. My assumption is that near-future additional ACS instances we plan to "own" will also be ACSCS-managed... as opposed to using self-installed RHACS installation we fully manage end-to-end ourselves on our clusters.

---

AppSRE's options seemingly are:

(A). We continue to use ACSCS. We use a separate app (e.g. QR) to manage ACS rbac via ACS-API.

(B). We continue to use ACSCS. We replace QR w/ our own hypothetical newly-built CRDs and k8s-controller to manage ACSCS instance RBAC in an Argo-native fashion.

(C). We continue to use ACSCS. We attempt to use [these patterns ](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_security_for_kubernetes/4.11/html/configuring/declarative-configuration-using ) to cluster-level-modify some of the saas-managed ACSCS instances on whatever cluster(s) our ACSCS Central instances are actually running on.

(D). We switch from ACSCS to RHACS. Then we can use these patterns [these patterns ](https://docs.redhat.com/en/documentation/red_hat_advanced_cluster_security_for_kubernetes/4.11/html/configuring/declarative-configuration-using ) for RBAC... but now we're newly in the business of managing the full lifecycle of ACS Central instances.

---

I assume we definitely don't desire to ever do (D)- otherwise AppSRE has to become experts in operating our own independent RHACS instances end-to-end. (Or perhaps you think AppSRE should switch from saas-managed ACSCS to self-managed RHACS, and that's the "disconnect" here).

I assume we probably don't desire (C) because cluster-level-manipulation of saas-managed services that we don't fully control sounds to me like an inherently bad/dangerous/complicated idea.

Maybe we could/should do (B), but then we'd be in the business of maintaining a new k8s-controller forever.

That seems to leave (A) as the least bad solution, even long term.

---

tl;dr I don't these docs are applicable to our situation. Or you have something else long-term in mind that I'm struggling to understand.

bkez322 (approved):
Happy to continue discussion if warranted, but this PR LGTM.

casey-williams-rh:
After discussing with @bkez322 and @aleccohan, we are going to move forward with the original QR implementation.

hemslo:
Asked in slack https://redhat-internal.slack.com/archives/C0313JYKH8W/p1786505535863999?thread_ts=1786505317.990219&cid=C0313JYKH8W, those features are not available for managed instances, so we still need to use ACS API/CLI to manage these resources.

PR #1050, opened 2026-08-12
[APPSRE-13941] Add automationTokens list fields to cluster schema

fltataru:
## Context

This is **Phase 1 of 5** in the `automationTokens` rotation migration.

> **Design doc:** [Vault Secret Rotation](https://github.com/openshift-online/platform-engineering-enhancements/blob/main/app-sre/design-docs/vault-secret-rotation.md)

### Migration plan

| Phase | Repo | Description | Status |
|-------|------|-------------|--------|
| 1 | qontract-schemas | Add `automationTokens`/`clusterAdminAutomationTokens` fields to cluster schema | **This MR** |
| 2 | qontract-reconcile | Prefer list tokens in `oc_connection_parameters`, implement list-based SA/Vault/MR workflow in `openshift-cluster-bots` | Pending |
| 3 | app-interface | Migrate clusters one-by-one: add `automationTokens` entries, let integration populate secrets, activate | Pending |
| 4 | qontract-schemas + qontract-reconcile | Remove singular fields from schema and all legacy code paths — this must happen together, as ~30 GQL queries across qontract-reconcile still reference the singular `automationToken`/`clusterAdminAutomationToken` fields | Pending |
| 5 | app-interface | Remove singular `automationToken`/`clusterAdminAutomationToken` from all clusters | Pending |

> **Note:** Phase 4 and the former Phase 5 are combined. The singular schema fields cannot be removed until all qontract-reconcile GQL queries referencing them have been updated in the same change.

### Design

The goal is to support **zero-downtime token rotation** for `openshift-cluster-bots`-managed clusters. Today each cluster has a single `automationToken` secret; rotating it requires a cutover window. The new model uses a list where multiple entries can coexist — the consumer picks the first `active: true, delete: false` entry with a populated secret — making rotations seamless.

Operators declare entries with `name`/`namespace`. The `openshift-cluster-bots` integration creates the ServiceAccount and Kubernetes secret, writes the token to Vault, and back-fills the `secret` reference via MR. The `active` flag signals readiness; `delete` flags entries for cleanup.

### Schema changes

- `automationTokens` — list of entries for the dedicated-admin SA token
- `clusterAdminAutomationTokens` — list of entries for the cluster-admin SA token
- `automationTokenEntry` defined inline in `cluster-1.yml` (not in `common-1.json` — it has no other consumers)
- An `allOf/not` constraint prevents `active: true` and `delete: true` from coexisting on the same entry
- `graphql-schemas/schema.yml` auto-updated by the schema generation tooling

## Test plan

- [ ] Schema validation passes for a cluster with `automationTokens` entries (covered by app-interface-dev-data test data)
- [ ] Schema validation rejects an entry with both `active: true` and `delete: true`
- [ ] Existing cluster files without `automationTokens` continue to validate (fields are optional)

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -430,6 +430,14 @@ confs:
   - { name: format, type: string }
   - { name: version, type: int }
 
+- name: AutomationTokenEntry_v1
+  fields:
+  - { name: name, type: string, isRequired: true }
+  - { name: namespace, type: string, isRequired: true }
+  - { name: active, type: boolean }
+  - { name: delete, type: boolean }
+  - { name: secret, type: VaultSecret_v1 }
+
 - name: KeyValue_v1
   fields:
   - { name: key, type: string, isRequired: true }
@@ -1154,8 +1162,10 @@ confs:
   - { name: insecureSkipTLSVerify, type: boolean }
   - { name: jumpHost, type: ClusterJumpHost_v1 }
   - { name: automationToken, type: VaultSecret_v1 }
+  - { name: automationTokens, type: AutomationTokenEntry_v1, isList: true }
   - { name: clusterAdmin, type: boolean }
   - { name: clusterAdminAutomationToken, type: VaultSecret_v1 }
+  - { name: clusterAdminAutomationTokens, type: AutomationTokenEntry_v1, isList: true }
   - { name: internal, type: boolean }
   - { name: disable, type: DisableClusterAutomations_v1 }
   - { name: awsInfrastructureAccess, type: AWSInfrastructureAccess_v1, isList: true }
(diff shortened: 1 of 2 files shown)


chassing (approved):

chassing:
Please be aware that other GQL queries, e.g. [namespaces minmal](https://github.com/app-sre/qontract-reconcile/blob/master/reconcile/gql_definitions/common/namespaces_minimal.gql), are using the singular versions too.
I think your step 4 must be after/together with step 5

fltataru:
Good catch, thanks! The description has been updated to address this — the former steps 4 and 5 are now combined into a single phase, with an explicit note that the singular schema fields cannot be removed until all ~30 qontract-reconcile GQL queries referencing them are updated in the same change.

PR #1062, opened 2026-08-31
[APPSRE-15181] add optional prefixStartSlaveCmd to Jenkins SSHConnector

gdragonetti:
What

The Jenkins EC2 Fleet worker fleets launch agents over SSH via the ssh-slaves plugin's SSHConnector. That connector supports a prefixStartSlaveCmd option that prepends a command to the agent launch. This change enables use of this field.

Why

We want to use this field to make the agent wait until the environment is ready.

### DIFF
--- a/graphql-schemas/schema.yml
+++ b/graphql-schemas/schema.yml
@@ -530,6 +530,7 @@ confs:
   - { name: launchTimeoutSeconds, type: int, isRequired: false }
   - { name: maxNumRetries, type: int, isRequired: false}
   - { name: port, type: int, isRequired: false}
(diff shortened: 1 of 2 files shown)


fishi0x01 (commented):
Great addition! However, we also need to add it here https://github.com/app-sre/qontract-schemas/blob/main/graphql-schemas/schema.yml

My recommendation would be to run integration locally with dry-run mode - e.g, with github.com/app-sre/qontract-development-cli. That way, you can test end-to-end if the integration runs and can properly query the data.

fishi0x01 (approved):
Good job! 🎉

PR #1072, opened 2026-09-10
[APPSRE-15234] Add managed-sso-client schema family

chassing:
## Summary

Adds a new schema family so tenants can declaratively request SSO clients on Keycloak directly from app-interface, instead of relying on the OCM-cluster-driven RHIDP flow or a manual SNOW ticket.

- `managed-sso-client-1.yml` - tenant-declared client, crossref'd to an `app` and a `keycloak-instance-1.yml`. `protocol` is a `oneOf` discriminator (mirroring `feature-toggle-1.yml`'s `provider` pattern) requiring `oidc` when set to `openid-connect`
- `keycloak-instance-1.yml` - reusable Keycloak instance/realm + IAT reference, letting tenants bring their own instance
- `managed-sso-client-oidc-1.yml` - full OIDC `ClientRepresentation` surface (~30 fields), each with a detailed description since a lot of Keycloak's OIDC behavior isn't obvious from field names alone
- `graphql-schemas/schema.yml`: `ManagedSsoClient_v1` is modeled as a proper interface (`interfaceResolve` on `protocol`) resolving to `ManagedSsoClientOpenidConnect_v1` - same shape as `FeatureToggle_v1` → `FeatureToggleUnleash_v1`. This gives client code a real discriminated Python type to `match` on instead of comparing the `protocol` string by hand

**Milestone 1 scope only**: `openid-connect` protocol. `saml` is reserved for a later milestone - adding it means one more `oneOf` branch and one more interface type, not a schema rework.

Design doc: https://github.com/openshift-online/platform-engineering-enhancements/pull/83
Jira: https://redhat.atlassian.net/browse/APPSRE-15234

Companion qontract-reconcile PR (Keycloak client extension + new `managed-sso-client` integration) to follow.

## Test plan

- [x] `uv run yamllint .` on the new/changed files
- [x] `make bundle` + `make validate` (via qontract-validator, `--only-errors` → `[]`)
- [x] Booted qontract-server against the bundled schema to confirm the GraphQL additions (including the new interface resolution) load cleanly with no schema errors
- [x] Regenerated the qontract-reconcile client-side query classes against this schema and confirmed `qenerate` produces the expected discriminated `ManagedSsoClientOpenidConnectV1` type


schemas/dependencies/keycloak-instance-1.yml
```diff
@@ -0,0 +1,37 @@
...
+    type: string
+    enum:
+    - /dependencies/keycloak-instance-1.yml
+  labels:
+    "$ref": "/common-1.json#/definitions/labels"
+  name:
+    "$ref": "/common-1.json#/definitions/identifierLowercase64"
+  description:
+    type: string
+  url:
```
hemslo, line 21: [P2] Restrict keycloakInstance.url to HTTPS

`format: uri` accepts `http://`, but the reconciliation client requires HTTPS and rejects those configurations at runtime. Add an HTTPS constraint so schema-valid configurations match the service contract.

chassing: Fixed in 820096c — added `pattern: "^https://"` to `keycloakInstance.url`.

hemslo (approved):
Re-reviewed current head fe96cfde. HTTPS validation is fixed, the nested OIDC schema no longer requires an inline schema marker, and CI is passing. No remaining actionable findings.

PR #1077, opened 2026-09-15
Add managed-sso-client openshift-resource provider

chassing:
## Summary
- New `managed-sso-client` provider in `schemas/openshift/openshift-resource-1.yml`: a namespace's `openshiftResources` entry can reference a `managed-sso-client-1.yml` object by `$ref` (via `managedSsoClient`) instead of hardcoding its tenant-facing Vault secret path via the generic `vault-secret` provider.
- New `NamespaceOpenshiftResourceManagedSsoClient_v1` GraphQL type + fieldMap entry in `graphql-schemas/schema.yml`.
- Registers both `managed-sso-client` and `openshift-managed-sso-client-secrets` as valid `disable.integrations` values in `schemas/openshift/cluster-integrations-1.yml`.

Implements Milestone 2 ("secret delivery") of the [managed-sso-client design doc](https://github.com/openshift-online/platform-engineering-enhancements/pull/83). Requires the paired [qontract-reconcile PR](https://github.com/app-sre/qontract-reconcile/pull/5821) implementing the `openshift-managed-sso-client-secrets` integration.

Jira: https://redhat.atlassian.net/browse/APPSRE-15234

## Test plan
- [x] Validated locally via `qontract-bundler` + `qontract-validator` docker images - zero validation errors
- [x] Validated the GraphQL schema loads cleanly via the `qontract-server` docker image against the bundled schema


schemas/openshift/openshift-resource-1.yml
```diff
@@ -145,6 +148,27 @@ oneOf:
   required:
   - path
   - type
+- additionalProperties: false
+  properties:
+    provider:
+      type: string
+      enum:
+      - managed-sso-client
```
hemslo, line 156: Could we avoid adding this temporary `managed-sso-client` provider and use the existing `vault-secret` contract for the bridge? A new provider creates a tenant-facing migration surface and risks having both qontract-reconcile and ESO own the same Secret. Please add it only if the existing provider cannot represent the required path, version, Secret name, or field mapping.

chassing: As discussed, we wanna keep a dedicated integration.

hemslo (approved):
LGTM!

